PR #10, opened 2023-11-30
ADR-794  Duty Suspended Deliveries Yes/No Page

Ewan-James-Donovan:
- Implemented a new Yes/No Page in line with the same patterns used by the draught relief question.
- Had some issues with GPG signing but I've retroactively resolved those, so please ignore relevant commits.
- Currently the page navigates to home and no navigation test has been implemented, as per ticket suggestion.
- There is one generated file I'm unsure whether to keep.
- The error message on the ticket didn't read well, so I did what I thought it should be:

Ticket says: 
`Select yes if this product if you need to declare and Duty Suspended Deliveries`
I put:
`Select yes if you need to declare Duty Suspended Deliveries`


app/controllers/DeclareDsdQuestionController.scala
```diff
@@ -0,0 +1,75 @@
...
+import navigation.Navigator
+import pages.DeclareDsdQuestionPage
+import play.api.i18n.{I18nSupport, MessagesApi}
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
+import views.html.DeclareDsdQuestionView
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DeclareDsdQuestionController @Inject() (
```
yasaracar, line 34: We are avoiding abbreviations in file names for better readibility, can you rename all the file name, vals, messages and routes to DeclareDutySuspendedDeliveries* please?

app/forms/DeclareDsdQuestionFormProvider.scala
```diff
@@ -0,0 +1,30 @@
...
+import javax.inject.Inject
+
+import forms.mappings.Mappings
+import play.api.data.Form
+
+class DeclareDsdQuestionFormProvider @Inject() extends Mappings {
+
+  def apply(): Form[Boolean] =
+    Form(
+      "value" -> boolean("declareDsdQuestion.error.required")
```
yasaracar, line 28: "value" need to be changed once you pull the id changes from the main into the branch

app/views/DeclareDsdQuestionView.scala.html
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @sectionHeading(
+        id = "declare-dsd-question-section",
+        text = messages("section.alcoholDutyReturn"),
+    )
+
+    @pageHeading(messages("declareDsdQuestion.heading"))
+
+    <p class="govuk-body govuk-!-margin-bottom-4">
```
yasaracar, line 46: Why do we use the margin bottom here?

migrations/DeclareDsdQuestion.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
yasaracar, line 1: Please run migrate.sh after you create a page using scaffold. migrate.sh will run these commands below(so it will create routes and messages for you) and then it will remove this file.
For this instance, please just remove this file (make sure you covered these commands manually)

conf/messages.en
```diff
@@ -90,3 +90,9 @@ alcoholByVolumeQuestion.error.minimumRequired = This product’s Alcohol by Volu
 alcoholByVolumeQuestion.error.maximumRequired = This product’s Alcohol by Volume (ABV) strength must be 100 or less
 alcoholByVolumeQuestion.error.twoDecimalPlaces = This product’s Alcohol by Volume (ABV) must be a number to two decimal places
 alcoholByVolumeQuestion.change.hidden = Alcohol by Volume (ABV) strength
+
+
+declareDsdQuestion.title = Do you need to declare alcohol that you delivered or received duty suspended?
+declareDsdQuestion.heading = Do you need to declare alcohol that you delivered or received duty suspended?
+declareDsdQuestion.error.required = Select yes if you need to declare Duty Suspended Deliveries
```
yasaracar, line 97: Please raise this with the Business Analyst and make sure we have the correct content

yasaracar (changes requested):

app/views/DeclareDsdQuestionView.scala.html
Ewan-James-Donovan: For some reason I needed this as the p tag was sitting too close to the radio buttons, but upon removing it I see that that problem, strangely,  is no longer there. We don't need this.

migrations/DeclareDsdQuestion.sh
Ewan-James-Donovan: Noted, I've covered it all manually

app/forms/DeclareDsdQuestionFormProvider.scala
Ewan-James-Donovan: I'll wait for those changes.

app/controllers/DeclareDsdQuestionController.scala
Ewan-James-Donovan: Done! :)

conf/messages.en
Ewan-James-Donovan: I'll do that now.

PR #12, opened 2023-12-06
Product Volume: Replace <p> with Hint + Remove "Question"

Ewan-James-Donovan:
As per Yash's suggestion, the p tag has been replaced with a govuk Hint component and the product volume page has been renamed


app/views/ProductVolumeQuestionView.scala.html
```diff
@@ -45,8 +45,6 @@
 
         @pageHeading(messages("productVolumeQuestion.heading"))
```
yasaracar, line 46: This can be removed now. Please remove the asVisuallyHidden() from the LabelViewModel below. The label will be header automatically then

yasaracar (changes requested):

yasaracar (approved):

PR #13, opened 2023-12-08
ADR-797 Duty Suspended Deliveries Inside UK input page

Ewan-James-Donovan:
Pretty much a carbon copy of the product volume page.

The prototype uses <p> and Hint - as such I've included both <p> and hint but not applied the same fixes that are in my other PR since they would render the page structure incorrect (the header would appear after the <p>)


test/forms/DutySuspendedDeliveriesFormProviderSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      fieldName,
+      minimum = minimum,
+      expectedError = FormError(fieldName, "dutySuspendedDeliveries.error.minimumRequired", ArraySeq(minimum))
+    )
+
+    behave like bigDecimalFieldWithMaximum(
+      form,
+      fieldName,
+      maximum = maximum,
+      expectedError = FormError(fieldName, "dutySuspendedDeliveries.error.maximumRequired", ArraySeq(999999999.99))
```
JemmaEdwards, line 61: we've done this in other spec files but why are we usng the value 999999999.99 and not the variable maximum?

Ewan-James-Donovan: all sorted

JemmaEdwards (dismissed):

JemmaEdwards (dismissed):

JemmaEdwards (approved):

PR #45, opened 2024-02-12
Adr 764: Product List screen

msatta:


app/controllers/productEntry/CheckYourAnswersController.scala
```diff
@@ -43,30 +46,57 @@ class CheckYourAnswersController @Inject() (
...
-
     }
+  }
 
+  private def getProductEntry(answers: UserAnswers, maybeInt: Option[Int]): Option[ProductEntry] = {
+    val pe = maybeInt match {
+      case Some(i) => answers.getByIndex(ProductEntryListPage, i)
+      case None    => answers.get(CurrentProductEntryPage)
+    }
+    pe.map(pe => if (pe.index.isDefined) pe else pe.copy(index = maybeInt))
```
yasaracar, line 82: Why do we do this? Isn't maybeInt None anyways if it is a CurrentProductEntry?

app/controllers/productEntry/CheckYourAnswersController.scala
```diff
@@ -43,30 +46,57 @@ class CheckYourAnswersController @Inject() (
...
   }
 
+  private def saveProductEntry(
+    answers: UserAnswers,
+    productEntry: ProductEntry
+  ): Future[UserAnswers] =
+    productEntry.index match {
+      case Some(i) => Future.fromTry(answers.setByIndex(ProductEntryListPage, i, productEntry.copy(index = None)))
```
yasaracar, line 90: Why do we set the index to None?

app/models/UserAnswers.scala
```diff
@@ -94,6 +94,16 @@ final case class UserAnswers(
       case JsError(errors)       =>
         Failure(JsResultException(errors))
     }
+
+  def setByIndex[A](page: Settable[Seq[A]], i: Int, value: A)(implicit writes: Writes[A]): Try[UserAnswers] = {
```
yasaracar, line 98: rename i to index please

app/navigation/ProductEntryNavigator.scala
```diff
@@ -69,4 +70,11 @@ class ProductEntryNavigator @Inject() () extends BaseNavigator {
       case Some(false) => routes.IndexController.onPageLoad
       case _           => routes.JourneyRecoveryController.onPageLoad()
     }
+
+  private def productListPageRoute(answers: UserAnswers): Call =
```
yasaracar, line 74: Test these please

yasaracar (changes requested):

app/controllers/productEntry/CheckYourAnswersController.scala
sini-george: maybeInt can have an index when user comes to the CYA page from clicking on 'change' on the Product List page. This index will be stored to the CurrentProductEntry. It helps when we need to decide whether we need to change a product entry in the list or the add a new one.

app/controllers/productEntry/CheckYourAnswersController.scala
sini-george: This is to clear the index we stored in the CurrentProductEntry when it comes from the 'Change' case described above

yasaracar (dismissed):

yasaracar (approved):

PR #67, opened 2024-04-09
ADR-1043 Update library versions and fix warnings

davidfes:


app/views/ViewUtils.scala
```diff
@@ -37,7 +37,7 @@ object ViewUtils {
     if (form.hasErrors || form.hasGlobalErrors) messages("error.browser.title.prefix") else ""
 
   def withPercentage(percentage: BigDecimal)(implicit messages: Messages) =
-    percentage + messages("site.unit.percentage")
```
davidfes, line 40: Believe this (warning) would also be fixed by percentage.toString + ... if + version was preferred.

davidfes: Yash okayed my version

yasaracar (approved):

PR #80, opened 2024-04-25
ADR-999 Add Other Spirits Produced page

davidfes:


app/viewmodels/govuk/TextAreaFluency.scala
```diff
@@ -0,0 +1,77 @@
...
+import uk.gov.hmrc.govukfrontend.views.viewmodels.hint.Hint
+import uk.gov.hmrc.govukfrontend.views.viewmodels.textarea.Textarea
+import uk.gov.hmrc.govukfrontend.views.viewmodels.label.Label
+import viewmodels.{ErrorMessageAwareness, InputWidth}
+
+object textarea extends TextareaFluency
+
+trait TextareaFluency {
+
+  object TextareaViewModel extends ErrorMessageAwareness {
```
yasaracar, line 30: Seems like we are using CharacterCountViewModel instead. Do we need to keep this?

app/views/spiritsQuestions/OtherSpiritsProducedView.scala.html
```diff
@@ -0,0 +1,63 @@
...
+            id = "other-spirits-produced-section",
+            text = messages("section.alcoholDutyQuarterlySpiritsReturn"),
+        )
+
+        @pageHeading(messages("otherSpiritsProduced.heading"))
+
+        @govukCharacterCount(
+            CharacterCountViewModel(
+                field = form("otherSpiritsProduced"),
+                label = LabelViewModel(messages("otherSpiritsProduced.subheading1")).asVisuallyHidden()
```
yasaracar, line 52: We use the input element's label as heading when there is no element between the heading and the first input element. Please see https://github.com/hmrc/alcohol-duty-returns-frontend/blob/main/app/views/productEntry/AlcoholByVolumeQuestionView.scala.html#L49

yasaracar (changes requested):

app/viewmodels/govuk/TextAreaFluency.scala
davidfes: We can delete it, but I wondered if it would be useful to leave it anyway to save that work being done again. Don't mind either way.

app/views/spiritsQuestions/OtherSpiritsProducedView.scala.html
davidfes: Fixed. Still waiting on a error message fix though before the ticket is done (moved back to blocked)

davidfes: Error messages updated.

yasaracar (dismissed):

yasaracar (approved):

PR #93, opened 2024-05-28
ADR-786 Check authorisation credentials

davidfes:


app/controllers/actions/AuthorisedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+    with FrontendHeaderCarrierProvider
+    with AuthorisedFunctions {
+
+  private val alcoholDutyEnrolment = "HMRC-AD-ORG"
+
+  override def invokeBlock[A](request: Request[A], block: Request[A] => Future[Result]): Future[Result] = {
+    implicit val headerCarrier: HeaderCarrier = hc(request)
+
+    authorised(
+      AuthProviders(GovernmentGateway)
```
yasaracar, line 50: We didn't need to create a new Action, we have the IdentifyAction for this. We are already authorising the users in the IdentifyAction, we just needed to change the code. Please move the relevant code to Identify Action and revert the changes in the controllers.

app/controllers/actions/AuthorisedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+class BaseAuthorisedAction @Inject() (
+  appConfig: FrontendAppConfig,
+  override val authConnector: AuthConnector,
+  val parser: BodyParsers.Default
+)(implicit val executionContext: ExecutionContext)
+    extends AuthorisedAction
+    with FrontendHeaderCarrierProvider
+    with AuthorisedFunctions {
+
+  private val alcoholDutyEnrolment = "HMRC-AD-ORG"
```
yasaracar, line 44: Use this from the config please: https://github.com/hmrc/alcohol-duty-returns-frontend/blob/main/conf/application.conf#L107
Anyways the IdentifyAction uses the config value when you move the code. Please see the comment below.

yasaracar (changes requested):

davidfes:
Given changing back the controllers would make a right mess of the commits, I'm going to close and reopen this as a new PR.

PR #94, opened 2024-05-28
ADR-786 Add authentication

davidfes:


davidfes:
Previous PR here: https://github.com/hmrc/alcohol-duty-returns-frontend/pull/93

test/controllers/actions/IdentifierActionSpec.scala
```diff
@@ -0,0 +1,232 @@
...
+              and CredentialStrength(strong)
+              and Organisation
+              and ConfidenceLevel.L50
+          ),
+          eqTo(
+            retriveInternalId and groupIdentifier and allEnrolments
+          )
+        )(any(), any())
+      )
+        .thenReturn(Future(combine(combine(Some(internalId), Some(groupId)), enrolments)))
```
davidfes, line 76: Couldn't work out why it wouldn't let me use ~ here so had to rename to combine. That said |+| and using cats Monoid would have been a better choice IMO but that's what it is.

yasaracar: Something I experienced before, it is very annoying. You can try one of these two approaches:

https://github.com/hmrc/ni-exemption-returns-frontend/blob/main/test/controllers/actions/auth/GGAuthActionSpec.scala#L198

https://github.com/hmrc/economic-crime-levy-returns-frontend/blob/main/test/uk/gov/hmrc/economiccrimelevyreturns/controllers/actions/AuthorisedActionSpec.scala#L53

yasaracar (changes requested):

app/controllers/actions/IdentifierAction.scala
```diff
@@ -58,10 +69,22 @@ class AuthenticatedIdentifierAction @Inject() (
       case _: NoActiveSession        =>
         Redirect(config.loginUrl, Map("continue" -> Seq(config.loginContinueUrl)))
       case _: AuthorisationException =>
+        handleAuthException
         Redirect(routes.UnauthorisedController.onPageLoad)
```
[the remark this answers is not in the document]
davidfes: Line 73 should be removed indeed since it's caught in the handleAuthException. Any other types aren't recovered as in the original. Good spot.

yasaracar (approved):

yasaracar (approved):

PR #96, opened 2024-05-30
ADR-1057: View past returns screen

sini-george:


app/controllers/ViewPastReturnsController.scala
```diff
@@ -0,0 +1,56 @@
...
+  val controllerComponents: MessagesControllerComponents,
+  view: ViewPastReturnsView,
+  alcoholDutyReturnsConnector: AlcoholDutyReturnsConnector
+)(implicit ec: ExecutionContext)
+    extends FrontendBaseController
+    with I18nSupport
+    with Logging {
+
+  def onPageLoad: Action[AnyContent] = identify.async { implicit request =>
+    (for { obligations <- alcoholDutyReturnsConnector.obligationDetails(request.appaId) } yield {
```
davidfes, line 45: As there is only one statement in the for, it could be replaced by:
alcoholDutyReturnsConnector.obligationDetails(request.appaId).map { obligations =>
...
}.recover...

app/viewmodels/ViewPastReturnsHelper.scala
```diff
@@ -0,0 +1,123 @@
...
+import play.twirl.api.Html
+import uk.gov.hmrc.govukfrontend.views.Aliases.{HeadCell, HtmlContent, Text}
+import uk.gov.hmrc.govukfrontend.views.html.components.{GovukTag, Tag}
+
+import java.time.format.DateTimeFormatter
+import java.time.{LocalDate, YearMonth}
+
+object ViewPastReturnsHelper extends Logging {
+
+  def outstandingReturnsTable(obligationData: Seq[ObligationData])(implicit messages: Messages): TableViewModel =
```
davidfes, line 31: This and the next method are the same, could they just be getReturnsTable or similar ?

app/viewmodels/ViewPastReturnsHelper.scala
```diff
@@ -0,0 +1,123 @@
...
+    messages: Messages
+  ): Html = {
+    val tag = status match {
+      case ObligationStatusToDisplay.Due       =>
+        Tag(content = Text(messages("viewPastReturns.status.due")), classes = "govuk-tag--blue")
+      case ObligationStatusToDisplay.Overdue   =>
+        Tag(content = Text(messages("viewPastReturns.status.overdue")), classes = "govuk-tag--red")
+      case ObligationStatusToDisplay.Completed =>
+        Tag(content = Text(messages("viewPastReturns.status.completed")), classes = "govuk-tag--green")
+      case _                                   => throw new RuntimeException("Couldn't create status for obligation data")
```
davidfes, line 109: Unnecessary as all 3 cases are covered above. The compiler will let you know if a new case was added you don't handle.

app/viewmodels/ViewPastReturnsHelper.scala
```diff
@@ -0,0 +1,123 @@
...
+    }
+    new GovukTag()(tag)
+  }
+
+  private def getPeriod(periodKey: String): YearMonth =
+    ReturnPeriod.fromPeriodKey(periodKey) match {
+      case Some(returnPeriod) => returnPeriod.period
+      case _                  =>
+        logger.logger.error("Couldn't fetch period from periodKey")
```
davidfes, line 118: Should this be a warn instead as it's not a wake someone up error, but more a problem with the data for this user.

app/views/ViewPastReturnsView.scala.html
```diff
@@ -0,0 +1,80 @@
...
+        @govukTable(Table(
+            caption = Some(messages("viewPastReturns.outstandingReturns.caption")),
+            captionClasses = "govuk-table__caption--m",
+            head = Some(outstandingReturns.head),
+            rows = outstandingReturns.rows.map(row => {
+                val labelRows = row.cells.map(cell => TableRow(cell))
+                val actions = TableRow(content = HtmlContent(tableRowActionList(
+                id = "outstandingReturnsId",
+                actions = row.actions.map(action => tableRowAction(
+                    id = "",
```
davidfes, line 49: Do we need an id here - same for table below?

test/connectors/AlcoholDutyReturnsConnectorSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+    ObligationStatus.Open,
+    LocalDate.of(2024, 1, 1),
+    LocalDate.of(2024, 1, 1),
+    LocalDate.of(2024, 1, 1),
+    periodKey
+  )
+  val appaIdentifier                          = appaIdGen.sample.get
+  val mockUrl                                 = s"http://alcohol-duty-returns/obligationDetails/$appaIdentifier"
+
+  "obligationDetails" - {
```
davidfes, line 51: This should really be tested with wireMock, but we can consider that tech debt.

test/viewmodels/ViewPastReturnsHelperSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+    "must return a table with the correct head" in {
+      val table = ViewPastReturnsHelper.outstandingReturnsTable(Seq(obligationDataSingleOpen))
+      table.head.size shouldBe 3
+    }
+
+    "must return a table with the correct rows" in {
+      val obligationData = Seq(obligationDataSingleOpen)
+      val table          = ViewPastReturnsHelper.outstandingReturnsTable(obligationData)
+      table.rows.size shouldBe obligationData.size
+      table.rows.map { case row =>
```
davidfes, line 65: case is unnecessary here since map takes a function not a partial function. Ditto for examples below.

davidfes: Note the confusion, if it's a tuple you want to decompose such as Seq[(String, String)] then it's .map { case (k, v) => .... }

app/viewmodels/ViewPastReturnsHelper.scala
sini-george: Good catch

app/views/ViewPastReturnsView.scala.html
sini-george: The view seems to fail when I take out the id

test/connectors/AlcoholDutyReturnsConnectorSpec.scala
davidfes: Tech debt ticket raised

test/viewmodels/ViewPastReturnsHelperSpec.scala
davidfes: Missed one, I've fixed it

davidfes (dismissed):

yasaracar (approved):

yasaracar (approved):

PR #97, opened 2024-05-31
ADR-1092 Add do you have an AppaId page

davidfes:
Requires 1097 to be merged first


davidfes:
See ADR-1092_3 for the new PR

PR #98, opened 2024-06-06
ADR-971 Check regimes for relevant DSD pages and for tasklist completion

davidfes:


app/controllers/TaskListController.scala
```diff
@@ -32,15 +32,24 @@ class TaskListController @Inject() (
...
+  view: TaskListView,
+  taskListService: TaskListService
 ) extends FrontendBaseController
     with I18nSupport
     with Logging {
 
   def onPageLoad: Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
     (request.userAnswers.validUntil, request.session.get(periodKeySessionKey)) match {
       case (Some(validUntil), Some(periodKey)) =>
-        Ok(view(AlcoholDutyTaskListHelper.getTaskList(request.userAnswers, validUntil, periodKey)))
```
davidfes, line 43: Using an object makes unit test hard, so changed to an injectable service

app/controllers/dutySuspended/DutySuspendedDeliveriesGuidanceController.scala
```diff
@@ -33,7 +38,11 @@ class DutySuspendedDeliveriesGuidanceController @Inject() (
...
     with I18nSupport {
 
-  def onPageLoad: Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
+  def onPageLoad(): Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
     Ok(view())
   }
+
+  def onSubmit(): Action[AnyContent] = (identify andThen getData andThen requireData) { request =>
+    Redirect(navigator.nextPage(DutySuspendedGuidancePage, NormalMode, request.userAnswers))
+  }
```
davidfes, line 47: So we can jump to the first relevant page via the navigator

app/models/RatePeriod.scala
```diff
@@ -54,47 +54,12 @@ object RateType {
     override def writes(o: RateType): JsValue = JsString(o.toString)
   }
 }
+
 case class RateTypeResponse(rateType: RateType)
 
 object RateTypeResponse {
   implicit val format: Format[RateTypeResponse] = Json.format[RateTypeResponse]
 }
-sealed trait AlcoholRegime
```
davidfes, line 62: Moved to own class using enumeratum

app/models/TaskListStatus.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package models
+
+import enumeratum.{Enum, EnumEntry}
+
+sealed trait TaskListStatus extends EnumEntry
+
+object TaskListStatus extends Enum[TaskListStatus] {
+  val values = findValues
+
```
davidfes, line 25: Preferred to strings (although converted to such to access messages)

app/models/package.scala
```diff
@@ -19,6 +19,7 @@ import play.api.libs.json._
 import scala.annotation.nowarn
 
 package object models {
+  type Result[T] = Either[Throwable, T]
```
davidfes, line 23: Saves having to write long form out everywhere we can return an error in an Either.

app/models/package.scala
```diff
@@ -19,6 +19,7 @@ import play.api.libs.json._
...
```
davidfes, line 22: Saves having to write out long form whenever we return an error via an Either

app/services/tasklist/TaskListService.scala
```diff
@@ -14,24 +14,30 @@
...
+import models.{Result, UserAnswers}
 import play.api.i18n.Messages
 import viewmodels.govuk.all.FluentInstant
+import viewmodels.tasklist.{AlcoholDutyTaskList, ReturnTaskListCreator}
 
 import java.time.Instant
+import javax.inject.Inject
 
-object AlcoholDutyTaskListHelper {
+class TaskListService @Inject() (returnTaskListCreator: ReturnTaskListCreator) {
```
davidfes, line 27: Using an injectable makes the unit testing easier.

app/viewmodels/tasklist/ReturnTaskListCreator.scala
```diff
@@ -26,50 +26,60 @@ import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
...
 import viewmodels.tasklist.DeclarationState.{Completed, InProgress, NotStarted}
 
-object ReturnTaskListHelper {
+import javax.inject.Inject
+
+class ReturnTaskListCreator @Inject() () {
   private def createSection(
     declareQuestionAnswer: Option[Boolean],
-    createTaskListSection: () => TaskListItem,
+    createTaskListSection: () => Either[Throwable, TaskListItem],
```
davidfes, line 34: Use Result[T] here ?

test/controllers/actions/CheckRegimeActionSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+  val userAnswersWithBeerRegime           = emptyUserAnswers.copy(data = JsObject(Seq("alcoholRegime" -> Json.toJson(Set(Beer)))))
+  val userAnswersWithCiderRegime          =
+    emptyUserAnswers.copy(data = JsObject(Seq("alcoholRegime" -> Json.toJson(Set(Cider)))))
+  val userAnswersWithWineRegime           = emptyUserAnswers.copy(data = JsObject(Seq("alcoholRegime" -> Json.toJson(Set(Wine)))))
+  val userAnswersWithSpiritsRegime        =
+    emptyUserAnswers.copy(data = JsObject(Seq("alcoholRegime" -> Json.toJson(Set(Spirits)))))
+  val userAnswersWithOtherFermentedRegime =
+    emptyUserAnswers.copy(data = JsObject(Seq("alcoholRegime" -> Json.toJson(Set(OtherFermentedProduct)))))
+
+  Seq(
```
davidfes, line 71: I'm not sure about the cache api, since get(page, json) returns a Try[JsObject] so you'd need to save the updated user answers at each stage to put in the next, hence I've used a foldLeft to act as a loop with state. Probably should look at it again, but since it only affects tests for now I've not made any change.

yasaracar (approved):

app/models/AlcoholRegimes.scala
```diff
@@ -0,0 +1,51 @@
...
+  def authorisedForRegime(regime: AlcoholRegime) =
+    if (regime == OtherFermentedProduct) {
+      hasOtherFermentedProduct()
+    } else {
+      hasRegime(regime)
+    }
+
+  private def hasRegime(regime: AlcoholRegime) = regimes.contains(regime)
+
+  def nextRegime(current: Option[AlcoholRegime]): Option[AlcoholRegime] = current match {
```
yasaracar, line 38: This looks to me a specific logic for Duty Suspended pages so better move it to DeclareDutySuspendedDeliveriesNavigator

app/models/TaskListStatus.scala
```diff
@@ -0,0 +1,28 @@
...
+  case object Completed extends TaskListStatus
```
yasaracar, line 26: Instead of calling Lowercase everytime this is used, please add entryName here

yasaracar (changes requested):

app/models/AlcoholRegimes.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import models.AlcoholRegime.{Beer, Cider, OtherFermentedProduct, Spirits, Wine}
+import play.api.libs.json.JsPath
+
+case class AlcoholRegimes(regimes: Set[AlcoholRegime]) {
+  def hasBeer()                  = regimes.contains(Beer)
+  def hasCider()                 = regimes.contains(Cider)
+  def hasWine()                  = regimes.contains(Wine)
+  def hasSpirits()               = regimes.contains(Spirits)
+  def hasOtherFermentedProduct() = hasCider() || hasWine() || regimes.contains(OtherFermentedProduct)
```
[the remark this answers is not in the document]
davidfes: Cider isn't add though which was one of the asks on the ticket to also go to OtherFermented. I agree if that was fixed, but is it safe to do so?

PR #102, opened 2024-06-21
ADR-971 Only show DSD pages if the appropriate regime is approved

davidfes:


app/controllers/TaskListController.scala
```diff
@@ -32,20 +32,16 @@ class TaskListController @Inject() (
...
   def onPageLoad: Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
     (request.userAnswers.validUntil, request.session.get(periodKeySessionKey)) match {
       case (Some(validUntil), Some(periodKey)) =>
-        Ok(
-          view(
-            AlcoholDutyTaskListHelper
-              .getTaskList(request.userAnswers, validUntil, periodKey)
-          )
-        )
+        Ok(view(taskListViewModel.getTaskList(request.userAnswers, validUntil, periodKey)))
```
davidfes, line 44: This is now injectable as an object made testing hard

app/models/TaskListStatus.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package models
+
+import enumeratum.{Enum, EnumEntry}
+
+sealed trait TaskListStatus extends EnumEntry {
+  def entryNameLower: String = entryName.toLowerCase
+}
+
+object TaskListStatus extends Enum[TaskListStatus] {
```
davidfes, line 25: Preferred enums to strings

app/controllers/TaskListController.scala
davidfes: Yash: I agree with making this injectable, thanks for the refactoring. But please move it back to viewmodels. We use service layer to orchestrate calls to backend microservices and for business logic. This class is more about view logic so better keeping it in viewmodels.

davidfes: Done

app/controllers/actions/CheckRegimeAction.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+trait CheckRegimeAction extends ActionRefiner[DataRequest, DataRequest] with Logging {
+  val regime: AlcoholRegime
+
+  override protected def refine[A](request: DataRequest[A]): Future[Either[Result, DataRequest[A]]] =
+    if (request.userAnswers.regimes.hasRegime(regime)) {
+      Future.successful(Right(request))
+    } else {
+      Future.successful(Left(Redirect(controllers.routes.UnauthorisedController.onPageLoad)))
```
davidfes, line 36: Yash:
As messaged earlier:

Shouldn't return Unauthorised but standard error page for now. We will use Unauthorised page for the cases where the user either needs to create a new GG account or should get an enrolment

app/controllers/actions/CheckRegimeAction.scala
```diff
@@ -0,0 +1,59 @@
...
+import play.api.mvc.Results.Redirect
+import play.api.mvc.{ActionRefiner, Result}
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+trait CheckRegimeAction extends ActionRefiner[DataRequest, DataRequest] with Logging {
+  val regime: AlcoholRegime
+
+  override protected def refine[A](request: DataRequest[A]): Future[Either[Result, DataRequest[A]]] =
```
davidfes, line 32: Yash:
As discussed earlier in a call, we are adding regimes: Seq[AlcoholRegime] in the DataRequest in ADR-1028 branch. Let's follow the same approach here

davidfes: That's merged now - the only thing that was relevant post-merge was view order - which I noted was defined by values in the enum. This (get the view order for what you have) was refactored and now is on AlcoholRegimes which navigator also uses now.

app/models/AlcoholRegimes.scala
```diff
@@ -27,6 +28,11 @@ case class AlcoholRegimes(regimes: Set[AlcoholRegime]) {
   def hasOtherFermentedProduct() = regimes.contains(OtherFermentedProduct)
```
davidfes, line 28: Yash: As discussed earlier in a call, we are adding regimes: Seq[AlcoholRegime] in the DataRequest in ADR-1028 branch. Let's follow the same approach here

davidfes: Cider isn't add though which was one of the asks on the ticket to also go to OtherFermented. I agree if that was fixed, but is it safe to do so?

davidfes: Update this was moved into account on 1132 (the pre-refactor) which due to misunderstandings isn't correct now. However we'd said we'd wait until outcomes of the regime approval discussion before making further changes. The auto-add of OFP from cider can be removed there. And in doing so a couple of the TestData examples here should be modified if changed (e.g. no Cider would now have OFP if not auto added) and maybe a couple of tests might need updating to cover that case.

app/models/AlcoholRegimes.scala
```diff
@@ -27,6 +28,11 @@ case class AlcoholRegimes(regimes: Set[AlcoholRegime]) {
...
 
   def hasRegime(regime: AlcoholRegime) = regimes.contains(regime)
+
+  def regimesInViewOrder(): Seq[AlcoholRegime] = AlcoholRegime.values.filter(regimes.contains)
+
+  def nextViewRegime(current: Option[AlcoholRegime]): Option[AlcoholRegime] =
+    current.fold(regimesInViewOrder().headOption)(regimesInViewOrder().toList.nextItem(_))
```
davidfes, line 35: Yash: This looks to me a specific logic for Duty Suspended pages so better move it to DeclareDutySuspendedDeliveriesNavigator

davidfes: Originally it got moved back there, but with regime view order being 'introduced' in 1028, I think this belongs here with the view order of regimes and may have other uses in the future. The navigator now doesn't have to know about view order (which is consistent across the server) as a concern, and avoids a potential duplication later.

app/models/TaskListStatus.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import enumeratum.{Enum, EnumEntry}
+
+sealed trait TaskListStatus extends EnumEntry {
+  def entryNameLower: String = entryName.toLowerCase
```
davidfes, line 22: Yash: Instead of calling Lowercase everytime this is used, please add entryName here

davidfes: IMO We shouldn't override entryName (since that's expected to return the enum as-is and if not final may have unintended consequences), but created a entryNameLower to cater for the one place we need to use it.

yasaracar (changes requested):

app/models/TaskListStatus.scala
yasaracar: I meant moving the whole TaskListStatus to ViewModels as they apply to views only but anyways we'll refactor later

yasaracar (approved):

PR #103, opened 2024-06-25
ADR-1079 Display specific return details

davidfes:


app/viewmodels/checkAnswers/returns/ViewReturnViewModel.scala
```diff
@@ -0,0 +1,181 @@
...
+      TableViewModel(head = noAdjustmentsTableHeader(), rows = nilAdjustmentsRow(), total = None)
+    }
+  }
+
+  private def adjustmentsTableHeader()(implicit messages: Messages): Seq[HeadCell] =
+    Seq(
+      HeadCell(
+        content = Text(messages("viewReturn.table.adjustmentType.legend"))
+      ),
+      HeadCell(content = Text(messages("viewReturn.table.description.legend"))),
```
davidfes, line 115: agreed not to create fifths or use quarters for this 5 column table (default behaviour seems to act on 20% spacing, but it's css so who knows what happens on other browsers/devices/platforms.

test/base/SpecBase.scala
```diff
@@ -47,12 +49,15 @@ trait SpecBase
     with IntegrationPatience
     with ModelGenerators
     with TestData {
-  def messages(app: Application): Messages = app.injector.instanceOf[MessagesApi].preferred(FakeRequest())
+  def getMessages(app: Application): Messages = app.injector.instanceOf[MessagesApi].preferred(FakeRequest())
```
davidfes, line 52: This was causing abiguity problems when naming the assignment val messages in certain cases, hence the rename rather than start writing msgs or something else to work around it.

test/base/SpecBase.scala
```diff
@@ -47,12 +49,15 @@ trait SpecBase
...
+
+  val fakeIdentifierUserDetails = FakeIdentifierUserDetails(appaId, groupId, internalId)
```
davidfes, line 54: Was using "appaId' as a the appaId which was causing issues when a generated one was being expected.

app/viewmodels/DateTimeHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package viewmodels
+
+import java.time.{Instant, LocalDateTime, YearMonth, ZoneId}
+import java.time.format.DateTimeFormatter
+import java.util.TimeZone
+import javax.inject.Inject
+
+class DateTimeHelper @Inject() () {
```
davidfes, line 24: Probably some places that can be 'fixed up' to use this and probably all date requests should come via here that need the correct zone, and maybe later localisation. However out of scope for now.

PR #105, opened 2024-06-26
ADR-1079/1080 Add link and ability to view a specific return (+ add logging on some failures)

davidfes:


app/controllers/actions/IndentifierWithoutEnrolment.scala
```diff
@@ -67,8 +67,10 @@ class AuthenticatedIdentifierWithoutEnrolmentAction @Inject() (
       block(IdentifierWithoutEnrolmentRequest(request, groupId, internalId))
     } recover {
       case e: AuthorisationException =>
+        logger.debug(s"Got AuthorisationException: $e")
         handleAuthException(e)
```
davidfes, line 71: 'm assuming we want these at debug level since it could spam the log in a DOS (however we still log other stuff anyway). Comments?

app/viewmodels/checkAnswers/productEntry/ProductListSummaryHelper.scala
```diff
@@ -35,7 +36,15 @@ object ProductListSummaryHelper {
         HeadCell(content = Text(messages("productEntryList.action")), classes = "govuk-!-width-one-quarter")
       ),
       rows = getProductEntryRows(productEntries),
-      total = productEntries.map(_.duty.getOrElse(BigDecimal(0))).sum
+      total = Some(
+        TableTotalViewModel(
+          HeadCell(content = Text(messages("productList.total")), classes = "govuk-!-width-one-half"),
+          HeadCell(
+            content = Text(messages("site.currency.2DP", productEntries.map(_.duty.getOrElse(BigDecimal(0))).sum)),
+            classes = "govuk-!-width-one-half"
```
yasaracar, line 44: Might be worth adding these to constants (applies to all class string values in the PR)

yasaracar (changes requested):

PR #113, opened 2024-07-31
ADR-1117 Fixed ViewReturnController not to use data retrieval actions

yasaracar:


TPilkington94Dev (approved):
Changes pushed by Yash, approved.

PR #116, opened 2024-08-02
ADR-1253: Adjustment under and over declaration reason with tasklist

sini-george:


app/viewmodels/tasklist/ReturnTaskListCreator.scala
```diff
@@ -270,13 +311,26 @@ class ReturnTaskListCreator @Inject() () {
...
+  def returnAdjustmentSection(userAnswers: UserAnswers)(implicit messages: Messages): Section = {
+    val overDeclarationTotal   = userAnswers.get(OverDeclarationTotalPage).getOrElse(BigDecimal(0))
+    val underDeclarationTotal  = userAnswers.get(UnderDeclarationTotalPage).getOrElse(BigDecimal(0))
+    val adjustmentListQuestion = userAnswers.get(AdjustmentListPage).getOrElse(true)
+    val taskListItems          = Seq(
+      Some(returnAdjustmentJourneyTaskListItem(userAnswers)),
+      if (underDeclarationTotal > 1000 && !adjustmentListQuestion) {
+        Some(returnAdjustmentJourneyUnderDeclarationTaskListItem(userAnswers))
+      } else { None },
+      if (overDeclarationTotal.abs > 1000 && !adjustmentListQuestion) {
```
yasaracar, line 323: Whay do we use abs here, but not in underDeclarationTotal?

app/controllers/adjustment/DeleteAdjustmentController.scala
```diff
@@ -60,12 +66,63 @@ class DeleteAdjustmentController @Inject() (
...
+                userAnswersWithUpdatedOverUnderReason <- updateOverUnderDeclarationReasons(updatedAnswers)
+                _                                     <- cacheConnector.set(userAnswersWithUpdatedOverUnderReason)
               } yield Redirect(controllers.adjustment.routes.AdjustmentListController.onPageLoad())
             } else {
               Future.successful(Redirect(controllers.adjustment.routes.AdjustmentListController.onPageLoad()))
             }
         )
   }
+
+  private def updateOverUnderDeclarationReasons(
```
yasaracar, line 79: same/similar logic is used in AdjustmentListController as well. Can they go in a helper class instead of duplicate code?

app/viewmodels/tasklist/ReturnTaskListCreator.scala
sini-george: as underDeclarationTotal will always have positive values, while overDeclarationTotal is always negative

yasaracar: Makes sense :)

yasaracar (dismissed):

yasaracar (approved):

PR #142, opened 2024-09-04
ADR-1307: Historical payments

sini-george:


app/controllers/returns/ViewPastPaymentsController.scala
```diff
@@ -44,29 +44,42 @@ class ViewPastPaymentsController @Inject() (
...
+          updatedSession,
+          outstandingPaymentsData.totalOpenPaymentsAmount
         )
       }
-      .recover { case _ =>
-        logger.warn(s"Unable to fetch open payments data for $appaId")
-        Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
+    // hardcoding 2024 until next iteration of past payments
+    val historicPaymentsFuture    =
+      alcoholDutyAccountConnector.historicPayments(appaId, 2024).map { historicPaymentsData =>
```
davidfes, line 67: Could this just be the current year? If we're going to do pagination shortly, then no problem.

davidfes (approved):

PR #144, opened 2024-09-04
ADR-1012 ContinueReturn Audit Event

JemmaEdwards:


app/controllers/BeforeStartReturnController.scala
```diff
@@ -75,4 +82,24 @@ class BeforeStartReturnController @Inject() (
...
+  )(implicit
+    hc: HeaderCarrier
+  ): Unit = {
+    val returnContinueTime = Some(Instant.now)
+    val eventDetail        = AuditContinueReturn(
+      appaId = returnId.appaId,
+      periodKey = returnId.periodKey,
+      governmentGatewayId = userAnswers.map(_.internalId),
+      governmentGatewayGroupId = userAnswers.map(_.groupId),
+      returnContinueTime = returnContinueTime,
```
yasaracar, line 98: why is this Optional? We have the value all the time so I think it should be mandatory

app/controllers/BeforeStartReturnController.scala
```diff
@@ -75,4 +82,24 @@ class BeforeStartReturnController @Inject() (
...
+      returnStartedTime = userAnswers.map(_.lastUpdated),
```
yasaracar, line 99: Can you comment on it with TODO to change it to start time in the future please?

test/controllers/BeforeStartReturnControllerSpec.scala
```diff
@@ -50,7 +56,10 @@ class BeforeStartReturnControllerSpec extends SpecBase {
         val result = route(application, request).value
 
         status(result) mustEqual SEE_OTHER
+
+        verify(mockAuditService).audit(any())(any(), any())
```
yasaracar, line 60: Might be worth changing these anys to eq to see if we audit the right values

yasaracar (changes requested):

yasaracar (approved):

PR #160, opened 2024-09-24
ADR-1221: View past returns hide table when no previous returns

sini-george:


app/views/returns/ViewPastReturnsView.scala.html
```diff
@@ -38,42 +39,50 @@
...
-                    text = action.label,
-                    href = action.href,
-                    visuallyHiddenContent = action.visuallyHiddenText
-                    )
-                ))))
-                row.cells :+ actions
-            }
-            ))
-        )
+        @if(outstandingReturns.rows.size > 0) {
```
davidfes, line 42: prefer .nonEmpty to size > 0

app/viewmodels/returns/ViewPastReturnsHelper.scala
```diff
@@ -35,11 +35,15 @@ class ViewPastReturnsHelper @Inject() () extends Logging {
...
-      total = None
-    )
+    if (sortedObligationData.nonEmpty) {
+      TableViewModel(
+        head = getTableHeader(messages),
+        rows = getObligationDataTableRows(sortedObligationData),
+        total = None
+      )
+    } else {
+      TableViewModel(head = Seq.empty, Seq.empty, total = None)
```
davidfes, line 45: Wonder if having a constructor method would be better here.

object TableViewModel {
  def empty(): TableViewModel = TableViewModel(Seq.empty, Seq,empty, None)
}

Also a bit inconsistent that the first and third parameter are named, but not the second.

app/views/returns/ViewPastReturnsView.scala.html
```diff
@@ -38,42 +39,50 @@
...
-                    text = action.label,
-                    href = action.href,
-                    visuallyHiddenContent = action.visuallyHiddenText
-                    )
-                ))))
-                row.cells :+ actions
-            }
-            ))
-        )
+        @if(completedReturns.rows.size > 0) {
```
davidfes, line 63: same comment here

test/viewmodels/checkAnswers/returns/ViewPastReturnsHelperSpec.scala
```diff
@@ -79,6 +86,17 @@ class ViewPastReturnsHelperSpec extends SpecBase with ScalaCheckPropertyChecks {
       }
     }
 
+    "must return the OverDue status an overdue obligation" in new SetUp {
+      val obligationData = Seq(obligationDataSingleOverDue)
+      val table          = viewPastReturnsHelper.getReturnsTable(obligationData)
+      table.rows.size shouldBe obligationData.size
+      table.rows.map { row =>
```
davidfes, line 93: since we don't need the result, prefer foreach to map

davidfes (approved):

test/viewmodels/returns/ViewPastReturnsHelperSpec.scala
```diff
@@ -86,17 +86,6 @@ class ViewPastReturnsHelperSpec extends SpecBase with ScalaCheckPropertyChecks {
       }
     }
 
-    "must return the OverDue status an overdue obligation" in new SetUp {
```
davidfes: As noted I've added this test (boundary check) in my unit test jira

davidfes (approved):

davidfes (approved):

msatta (approved):

PR #168, opened 2024-10-01
ADR-1352 before you start additional content

JemmaEdwards:


yasaracar (changes requested):

app/views/BeforeStartReturnView.scala.html
```diff
@@ -54,6 +56,27 @@
 
         @paragraph(messages("beforeStartReturn.text.guidance4"))
 
+        @if(currentDate.isAfter(returnDueDate)) {
```
yasaracar, line 59: Change the whole clock to if else please:
if after ->
else if before -> 
else

yasaracar (changes requested):

app/viewmodels/WarningTextViewModel.scala
```diff
@@ -0,0 +1,44 @@
...
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+import uk.gov.hmrc.govukfrontend.views.viewmodels.warningtext.WarningText
+import viewmodels.returns.ReturnPeriodViewModel.viewDateFormatter
+
+import java.time.LocalDate
+
+case class WarningTextViewModel(returnPeriod: ReturnPeriod, currentDate: LocalDate) {
+  private val returnDueDate                                  = returnPeriod.periodDueDate()
+  def getDateValue(implicit messages: Messages): WarningText =
```
yasaracar, line 29: This returns a warnintText, so rename getDateValue to warningText

app/viewmodels/WarningTextViewModel.scala
```diff
@@ -0,0 +1,44 @@
...
+    if (currentDate.isBefore(returnDueDate)) {
+      WarningText(
+        iconFallbackText = Some("Warning"),
+        content = Text(messages("beforeStartReturn.text.dueDateWarning", viewDateFormatter.format(returnDueDate)))
+      )
+    } else if (currentDate.isAfter(returnDueDate)) {
+      WarningText(
```
yasaracar, line 36: To much code duplication. Create a val for the messages which is decided by an if and use that val once in the WarningText

yasaracar (changes requested):

yasaracar (approved):

PR #203, opened 2024-11-06
ADR-1456 custom login page

simondrugan16:
Attempting to create a custom login page so that we can bypass the auth wizard easily, making the service more accessible to stakeholders and non technical people


simondrugan16:
Low priority. When picked up again will be better to start from the start :')

PR #214, opened 2024-11-19
ADR-1466: Spoilt duty rate displayed as NA on view specific return

sini-george:


app/viewmodels/returns/ViewReturnViewModel.scala
```diff
@@ -187,30 +187,32 @@ class ViewReturnViewModel @Inject() (appConfig: FrontendAppConfig) {
...
-      val maybeRatePeriod = ReturnPeriod.fromPeriodKey(returnAdjustmentsRow.returnPeriodAffected).map(_.period)
-      val taxType         = returnAdjustmentsRow.taxType
-      val description     = if (returnAdjustmentsRow.adjustmentTypeKey.equals(ReturnAdjustments.spoiltKey)) {
-        appConfig.getRegimeNameByTaxTypeCode(taxType) match {
-          case Some(regime) => messages(s"alcoholType.$regime")
-          case _            => taxType
-        }
+      val maybeRatePeriod         = ReturnPeriod.fromPeriodKey(returnAdjustmentsRow.returnPeriodAffected).map(_.period)
+      val taxType                 = returnAdjustmentsRow.taxType
+      val (description, dutyRate) = if (returnAdjustmentsRow.adjustmentTypeKey.equals(ReturnAdjustments.spoiltKey)) {
```
simondrugan16, line 192: I think it would be clearer if the description and dutyRate were in separate vals. Then you could extract
`returnAdjustmentsRow.adjustmentTypeKey.equals(ReturnAdjustments.spoiltKey)`
as a private def called isSpoiltRow or something. I think this would be easier to read like this

conf/messages.en
```diff
@@ -150,6 +150,7 @@ viewReturn.table.dutyDue.legend = Duty value
 viewReturn.printYourReturn = Print your return
 viewReturn.netDutySuspension.caption = Duty suspended deliveries
 viewReturn.netDutySuspension.noneDeclared = Nothing declared
+viewReturn.notApplicable = NA
```
simondrugan16, line 153: Can you check with Del that this should be NA rather than N/A. N/A is a pretty standard abbreviation of Not Applicable even though the ticket says to do NA

app/viewmodels/returns/ViewReturnViewModel.scala
sini-george: Thank you for the suggestion.. I see how separating `description` and `dutyRate` and extracting isSpoiltRow  as a separate def could improve readability in this case. However, I believe keeping description and dutyRate together in a single val destructuring better reflects the relationship between them, as they both change only when the adjustment is of type spoilt. By handling this logic in one place, the code remains cohesive and avoids scattering related computations, which I think makes it easier to follow. Does this approach sound fair to you?

simondrugan16: Very fair 😊 I'll mark this as resolved

simondrugan16 (dismissed):

simondrugan16 (approved):

PR #215, opened 2024-11-19
ADR-1455 fix spirits navigation

simondrugan16:
There was a bug where the declare quarterly spirits page would error out when selecting a new option for Yes/No (AKA in change mode). I fixed this by using User Answers to decide where to redirect to based on the current answer to the question


simondrugan16:
closing since i merged this pr into https://github.com/hmrc/alcohol-duty-returns-frontend/pull/204

PR #218, opened 2024-11-25
ADR-1518 Display charge reference on view return

davidfes:


test/controllers/returns/ViewReturnControllerSpec.scala
```diff
@@ -124,6 +125,7 @@ class ViewReturnControllerSpec extends SpecBase {
         status(result) mustEqual OK
         contentAsString(result) mustEqual view(
           returnPeriodStr,
+          Some(chargeReference),
```
sini-george, line 128: Can we have a view which has None charge reference case in one of these tests or a new test

sini-george (approved):

sini-george (dismissed):

sini-george (approved):

PR #222, opened 2024-11-27
ADR 1478 add class and method to loggers

simondrugan16:
Making the message field in logs more helpful when scanning. The Class and Method name where the logger originated from will show in the message which will hopefully be helpful!

### DIFF
--- a/app/controllers/BeforeStartReturnController.scala
+++ b/app/controllers/BeforeStartReturnController.scala
@@ -93,7 +93,7 @@ class BeforeStartReturnController @Inject() (
             auditReturnStarted(userAnswer)
             Redirect(controllers.routes.TaskListController.onPageLoad)
           case Left(error)       =>
-            logger.warn(s"Unable to create userAnswers:", error)
+            logger.warn(s"Unable to create userAnswers: $error")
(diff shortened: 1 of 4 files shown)


ajchilts (approved):
Looks good to me @simondrugan16  . Do we have an example of the output we could attach to the PR please? Before and after maybe? 

Nice. 👍

simondrugan16:
Example (truncated some of the stuff that stayed the same):
Old:
`level=[WARN] message=[Unable to create userAnswers: uk.gov.hmrc.http.UpstreamErrorResponse: POST of 'http://localhost:16001/alcohol-duty-returns/user-answers' returned 404. Response body: '{"statusCode":404,"message":"Entity not found"}'] `
New:
`level=[WARN] message=[BeforeStartReturnController|$anonfun$onSubmit$3|96 - Unable to create userAnswers: uk.gov.hmrc.http.UpstreamErrorResponse: POST of 'http://localhost:16001/alcohol-duty-returns/user-answers' returned 404. Response body: '{"statusCode":404,"message":"Entity not found"}'] `

The new version tells us the class, method and line where the logger was made so that it's easier to see where this error is. I'm hoping these will show up on Kibana logs too with the changes in the ""  appender name="FILE" "" bit

PR #225, opened 2024-11-29
Revert "ADR 1478 add class and method to loggers"

simondrugan16:
Reverts hmrc/alcohol-duty-returns-frontend#222


simondrugan16:
Reverting since doesn't have any impact on Kibana logs

ajchilts (approved):
👍

simondrugan16:
Closing because it seems like this actually is working in Kibana :)

PR #227, opened 2024-12-02
ADR-1448 formatter improvements

sini-george:


ajchilts (commented):
We would need to test each affected page and ensure that the order of validation is upheld however, I do like this change as its less cognitive load on the user. 👍

ajchilts (approved):
LGTM 👍

simondrugan16 (approved):

PR #232, opened 2024-12-05
ADR-1460 

simondrugan16:
Ticket [ADR-1460](https://jira.tools.tax.service.gov.uk/browse/ADR-1460). What I did is create a new action that checks if the user is registered as an organisation. If this is the case then when clicking the sign out button, everything works the same as it currently is. If the user is NOT an organisation, then we will not try and identify them, and just redirect them to the sign out url without trying to remove the lock on their account (because in this case they would not have the lock, because they are not signed in and do not have associated AppaIds). Apart from this, the functionality is the same.


simondrugan16:
impl done wrong (not checking enough perquisites. new PR checks that all auth is done before releasing the lock: https://github.com/hmrc/alcohol-duty-returns-frontend/pull/236

PR #235, opened 2024-12-05
ADR-1535 Select appropriate regime where possible in dynamic content

davidfes:


ajchilts (dismissed):

app/viewmodels/declareDuty/RateBandHelper.scala
```diff
@@ -16,18 +16,28 @@
...
 import play.api.i18n.Messages
 
 object RateBandHelper {
 
-  def rateBandContent(rateBand: RateBand)(implicit messages: Messages): String                         =
-    rateBand.rangeDetails.flatMap(_.abvRanges.toSeq) match {
+  def rateBandContent(rateBand: RateBand, maybeByRegime: Option[AlcoholRegime])(implicit
+    messages: Messages
+  ): String = {
+    val rangeDetails =
```
yasaracar, line 27: This is repeated below, can we create a function like abvRangesFromRateBand?

yasaracar (changes requested):

app/viewmodels/declareDuty/RateBandHelper.scala
davidfes: Done, makes sense.

yasaracar (approved):

PR #236, opened 2024-12-06
ADR-1460-v2

simondrugan16:
Ticket [ADR-1460](https://jira.tools.tax.service.gov.uk/browse/ADR-1460). What I did is create a new action that checks if the user is fully authorised. If this is the case then when clicking the sign out button, everything works the same as it currently is. If the user is NOT authorised, then that's ok, we just redirect them to the sign out url without trying to remove the lock on their account (because in this case they would not have the lock, because they are not signed in and do not have associated AppaIds). Apart from this, the functionality is the same.


simondrugan16:
Slight risk on this ticket. If a user is on the return locked page when they sign out they are fully authenticated. The release lock fails because a Period key cannot be found in their session. Need to think on this

app/config/Module.scala
```diff
@@ -27,6 +27,7 @@ class Module extends AbstractModule {
 
     bind(classOf[DataRetrievalAction]).to(classOf[DataRetrievalActionImpl]).asEagerSingleton()
     bind(classOf[DataRequiredAction]).to(classOf[DataRequiredActionImpl]).asEagerSingleton()
+    bind(classOf[SignOutAction]).to(classOf[SignOutActionImpl]).asEagerSingleton()
```
ajchilts, line 30: We went for the action. 👍

app/controllers/actions/SignOutAction.scala
```diff
@@ -0,0 +1,87 @@
...
+      val appaId = getAppaId(enrolments)
+      block(RequestWithOptAppaId(request, Some(appaId)))
+    } recoverWith {
+      case e: AuthorisationException =>
+        logger.debug(s"Returning a request with AppaId set to None since there was an AuthorisationException:", e)
+        block(RequestWithOptAppaId(request, None))
+      case e: UnauthorizedException  =>
+        logger.debug(s"Returning a request with AppaId set to None since there was an UnauthorizedException:", e)
+        block(RequestWithOptAppaId(request, None))
+    }
```
ajchilts, line 69: Do we need to distinguish between these two exception types if the outcome is the same? The logout will show the exception type in KIbana? Combine to a single case?

app/controllers/auth/AuthController.scala
```diff
@@ -33,22 +33,29 @@ class AuthController @Inject() (
...
+            userAnswersConnector
+              .releaseLock(ReturnId(appaId, periodKey))
+              .map(_ => Redirect(config.signOutUrl, Map("continue" -> Seq(config.exitSurveyUrl))))
+          case None            =>
+            logger.info("Period key not found during sign out")
+            Future.successful(Redirect(config.signOutUrl, Map("continue" -> Seq(config.exitSurveyUrl))))
+        }
+      case None         =>
+        logger.info(
+          "AppaId was None in the request which means that this user was not authorised. No locks on their account have been released."
```
ajchilts, line 56: This logout is making an assumption. Maybe it's ok to have a generic message that states the user was not authenticated. No locks to release?

app/controllers/actions/SignOutAction.scala
simondrugan16: There are different types of exceptions from AuthConnector which will trigger these different cases. E.g: unsupported affinity group, e.g: if the user is not an organisation (which is what I tunnel visioned on in the first PR 😅), then we will hit that first case AuthorisationException. Other errors like can't get confidence level/confidence level not 50 then we get the second case, UnauthorizedException.

AFAIK you can't do a double type check in one case statement without something like:
  case e if e.isInstanceOf[AuthorisationException] || e.isInstanceOf[UnauthorizedException] =>
which is more complex than the original IMO

All other exceptions we don't want to catch really, since the standard error page is ideal (issue with Auth or something). On the other Actions we would redirect to the UnauthorisedController in this case which would take us back to the loop we had before :D

ajchilts (approved):

yasaracar (approved):

ajchilts (approved):

PR #242, opened 2024-12-13
ADR-1192 Display Tax type code description as a key for declare duty summary cards

davidfes:


ajchilts (approved):
LGTM 👍

ajchilts (dismissed):
LGTM 👍

ajchilts (approved):
👍 looks like this aligns with the spec more now. Nice.

I think if there is any UI (css) required we could do that as a separate design ticket if we need to progress these changes first?

PR #249, opened 2024-12-19
ADR-1587 Prevent accessing submission page or submitting if tasks are incomplete

davidfes:


app/controllers/checkAndSubmit/DutyDueForThisReturnController.scala
```diff
@@ -57,24 +60,50 @@ class DutyDueForThisReturnController @Inject() (
...
+        } else {
+          EitherT.leftT[Future, Unit](
+            s"Check and submit page reached without completing all tasks ${userAnswers.returnId.appaId}/${userAnswers.returnId.periodKey}"
+          )
+        }
+      dutyDueViewModel <- dutyDueForThisReturnHelper.getDutyDueViewModel(userAnswers, returnPeriod)
+    } yield dutyDueViewModel
+
+    dutyDueViewModelOrError.foldF(
+      error => {
```
ajchilts, line 79: I assume this foldF will yield the error from line 72 should it have occurred? I.e. if the task list is incomplete?

app/controllers/checkAndSubmit/DutyDueForThisReturnController.scala
```diff
@@ -57,24 +60,50 @@ class DutyDueForThisReturnController @Inject() (
...
+    val userAnswers  = request.userAnswers
+    val returnPeriod = request.returnPeriod
+    val appaId       = userAnswers.returnId.appaId
+    val periodKey    = userAnswers.returnId.periodKey
+
     val result = for {
+      // Recheck we have everything
+      // The user can have a duplicate tab open, change stuff then use it to submit
+      _                           <-
+        if (taskListViewModel.checkAllDeclarationSectionsCompleted(userAnswers, returnPeriod)) {
```
ajchilts, line 97: If we are repeating this block could we refactor into a function?

ajchilts (commented):
Couple of minor questions @davidfes .

app/controllers/checkAndSubmit/DutyDueForThisReturnController.scala
davidfes: yes, so the for comprehension returns an EitherT[Future, String, DutyDueViewModel] - so it'll either execute the first function if it's a string (error), or the second if it's a DDVM (successful). The for comprehension bails out with the error if any statement returns a Future.failed or Future.successful(Left(xxx))

app/controllers/checkAndSubmit/DutyDueForThisReturnController.scala
davidfes: Perhaps. The error itself is different. While it's a very functional thing to do, I think it can make it less readable what's going on (we'd need to pass the error in as well). For two instances in the same file I thought it would not give benefits, but happy to do it though if you prefer it.

ajchilts (approved):
LGTM 👍

PR #250, opened 2024-12-19
ADR-1571 Implement validations mandating 4 decimals in LPA fields for R&A

Bradleigh-L:


.g8/multipleQuestionsPage/generated-test/controllers/$section$/$className$ControllerSpec.scala
```diff
@@ -17,7 +17,7 @@ import connectors.UserAnswersConnector
 import uk.gov.hmrc.http.HttpResponse
 import views.html.$section$.$className$View
 
-import scala.concurrent.Future
```
davidfes, line 20: shouldn't have checked in

.g8/radioButtonPage/generated-test/controllers/$section$/$className$ControllerSpec.scala
```diff
@@ -16,7 +16,7 @@ import connectors.UserAnswersConnector
 import uk.gov.hmrc.http.HttpResponse
 import views.html.$section$.$className$View
 
-import scala.concurrent.Future
```
davidfes, line 19: shouldn't have checked in

app/forms/mappings/AdjustmentVolumesAndRateFormatter.scala
```diff
@@ -32,50 +33,51 @@ class AdjustmentVolumesAndRateFormatter(
...
-    fieldKey,
+    sprDutyRateField,
     maximumValue = Constants.dutyMaximumValue,
     minimumValue = Constants.dutyMinimumValue,
     args = args
   )
 
   private def requiredFieldFormError(key: String, field: String): FormError =
-    FormError(nameToId(s"${key}_$field"), s"$requiredKey.$field", args)
+    FormError(nameToId(s"$key.$field"), s"$requiredKey.$field", args)
```
davidfes, line 75: name to id already changes . to _  so no need to do it here - hence the change.

app/forms/mappings/AdjustmentVolumesAndRateFormatter.scala
```diff
@@ -113,22 +115,19 @@ class AdjustmentVolumesAndRateFormatter(
     }
   }
 
-  private def validateField(
+  private[mappings] def validateField[T](
```
davidfes, line 118: Q do we need the package here?

app/forms/mappings/AdjustmentVolumesFormatter.scala
```diff
@@ -96,20 +98,18 @@ class AdjustmentVolumesFormatter(
     }
   }
 
-  private def validateField(
+  private[mappings] def validateField[T](
```
davidfes, line 101: same Q about the package here

app/forms/mappings/BigDecimalFieldFormatter.scala
```diff
@@ -62,5 +68,12 @@ class BigDecimalFieldFormatter(
       }
 
   override def unbind(key: String, value: BigDecimal): Map[String, String] =
-    baseFormatter.unbind(key, value.toString)
+    if (exactDecimalPlacesRequired) {
+      baseFormatter.unbind(
+        key,
+        value.setScale(decimalPlaces).toString
+      ) // TODO could we use the play implementation to do a tostring here?
```
davidfes, line 75: remove TODO

app/forms/mappings/SpoiltVolumesAndDutyFormatter.scala
```diff
@@ -94,40 +96,38 @@ class SpoiltVolumesAndDutyFormatter(
...
+      totalLitresResult.left.toSeq.flatten ++ pureAlcoholResult.left.toSeq.flatten ++ dutyRateResult.left.toSeq.flatten
     if (allErrors.nonEmpty) {
       Left(allErrors)
     } else {
       checkValues(key, data)
     }
   }
 
-  private def validateField(
+  private[mappings] def validateField[T](
```
davidfes, line 118: same Q about the package

app/viewmodels/tasklist/AlcoholDutyTaskListItemStatus.scala
```diff
@@ -26,19 +25,19 @@ import uk.gov.hmrc.govukfrontend.views.viewmodels.tasklist.TaskListItemStatus
 object AlcoholDutyTaskListItemStatus {
   def completed(implicit messages: Messages): TaskListItemStatus  = TaskListItemStatus(
     content = Text(messages("taskList.section.status.completed")),
-    classes = Constants.textAlignRightCssClass
+    classes = Css.textAlignRightCssClass
```
davidfes, line 28: was some clean up in Constants - this should use the Css object

test/forms/mappings/MappingsSpec.scala
```diff
@@ -789,13 +807,51 @@ class MappingsSpec extends SpecBase with Mappings {
       }
     }
 
+    "must not bind pureAlcohol values" - {
```
davidfes, line 810: Need these tests in the other formatters

.g8/radioButtonPage/generated-test/controllers/$section$/$className$ControllerSpec.scala
Bradleigh-L: Removed

app/forms/mappings/AdjustmentVolumesAndRateFormatter.scala
Bradleigh-L: Reverted

conf/messages.en
```diff
@@ -819,9 +819,9 @@ adjustmentVolume.error.maximumValue.sprDutyRate = Your Small Producer Relief dut
 adjustmentVolume.error.decimalPlaces.sprDutyRate = Your Small Producer Relief duty rate must be to 2 decimal places
 adjustmentVolume.change.hidden = volume
 adjustmentVolume.totalLitresVolume.heading = Total litres
-adjustmentVolume.totalLitresVolume.hint = In litres, up to 2 decimal places
+adjustmentVolume.totalLitresVolume.hint = To 2 decimal places
```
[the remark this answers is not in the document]
davidfes: Fixed

app/forms/mappings/AdjustmentVolumesFormatter.scala
```diff
@@ -96,20 +98,18 @@ class AdjustmentVolumesFormatter(
     }
   }
 
-  private def validateField(
+  private def validateField[T](
```
simondrugan16, line 101: why make this generic by adding a type param when we only use BigDecimalFieldFormatter?

app/forms/mappings/BigDecimalFieldFormatter.scala
```diff
@@ -32,11 +32,17 @@ class BigDecimalFieldFormatter(
   args: Seq[String] = Seq.empty,
   decimalPlaces: Int = Constants.maximumDecimalPlaces,
```
simondrugan16, line 33: I think Constants.maximumDecimalPlaces should be renamed to something like Constants.decimalDecimalPlaces, because it's not the maximum number of Decimal Places

app/forms/mappings/BigDecimalFieldFormatter.scala
```diff
@@ -62,5 +68,12 @@ class BigDecimalFieldFormatter(
...
 
   override def unbind(key: String, value: BigDecimal): Map[String, String] =
-    baseFormatter.unbind(key, value.toString)
+    if (exactDecimalPlacesRequired) {
+      baseFormatter.unbind(
+        key,
+        value.setScale(decimalPlaces).toString
+      )
+    } else {
+      baseFormatter.unbind(key, value.toString)
```
simondrugan16, line 77: if, for example, someone enters 9.10 and exactDecimalPlacesRequired is false, will this get trimmed to 9.1 when the page reloads?

conf/messages.en
```diff
@@ -810,7 +810,7 @@ adjustmentVolume.change.hidden = volume
 adjustmentVolume.totalLitresVolume.heading = Total litres
 adjustmentVolume.totalLitresVolume.hint = In litres, up to 2 decimal places
 adjustmentVolume.pureAlcoholVolume.heading = Litres of pure alcohol
-adjustmentVolume.pureAlcoholVolume.hint = In litres, up to 4 decimal places
+adjustmentVolume.pureAlcoholVolume.hint = To 4 decimal places
```
simondrugan16, line 813: Ticket says to change "Up to 4 decimal places" --> "To 4 decimal places". It doesn't mention the "In litres" bit. Are you sure this is correct?

test/forms/mappings/MappingsSpec.scala
```diff
@@ -407,10 +407,45 @@ class MappingsSpec extends SpecBase with Mappings {
       }
     }
 
+    "must not bind pureAlcoholVolume values" - {
```
simondrugan16, line 410: do we need a test for binding when there are more than 4 dps?

test/forms/mappings/MappingsSpec.scala
```diff
@@ -1093,4 +1257,310 @@ class MappingsSpec extends SpecBase with Mappings {
...
+      "decimalPlaces",
+      "minimumValue",
+      "maximumValue",
+      "inconsistentKey",
+      Seq.empty
+    )
+
+    val testForm: Form[SpoiltVolumeWithDuty] = Form("value" -> mapping)
+
+    "must bind a validSpoiltVolumeWithDuty" in {
```
simondrugan16, line 1274: space between valid + SpoiltVolumeWithDuty?

test/forms/mappings/MappingsSpec.scala
```diff
@@ -1093,4 +1257,310 @@ class MappingsSpec extends SpecBase with Mappings {
...
+          Map(
+            "value.totalLitresVolume" -> "1234.45",
+            "value.pureAlcoholVolume" -> "12.345",
+            "value.duty"              -> "5.31"
+          )
+        )
+        result.errors mustBe Seq(FormError("value_pureAlcoholVolume", "decimalPlaces.pureAlcoholVolume", Seq.empty))
+      }
+
+      "with no decimal places and trailing decimal point" in {
```
simondrugan16, line 1402: is there a test for value.totalLitresVolume with a trailing dot?

app/forms/mappings/BigDecimalFieldFormatter.scala
davidfes: Yes I believe that's the case right now and we fixed that for the 4dp ones. Mongo doesn't store the trailing places. The problem is we wouldn't know if they really entered 9.1 or 9.10 - I guess the question is whether it matters.

davidfes: Raised the question with Tom. I don't think it matters since we would expect if users have a non-zero in the second decimal place they enter it rather than round, thus 9.1 = 9.10.

conf/messages.en
davidfes: Raised with Tom. Confluence was saying to remove the litres. I guess that's because it's already in the label of the entry field.

test/forms/mappings/MappingsSpec.scala
davidfes: I'll check, I believe there was such a test in all places.

test/forms/mappings/MappingsSpec.scala
davidfes: yes, good spot.

test/forms/mappings/MappingsSpec.scala
davidfes: I don't think there is. It's low risk for me to add one. It's minor though since I'm not expecting users to enter stuff like that and the regex is the only reason it can be entered that way. I guess a test would show it's not doing anything unexpected.

davidfes: I'll check this with Tom whether we want to continue to allow this. For me it looks like they had decimal places but failed to submit them and personally I'd disallow it totally. Any change here rather than a test for good citizen sake is outside the scope of the ticket, though was discovered as a possibility during work on this ticket.

Any change to disallow a trailing decimal point would have to be a separate ticket.

app/forms/mappings/AdjustmentVolumesFormatter.scala
davidfes: I think it was to be consistent across formatters which I think is reasonable. I'll rereview why this was done.

app/forms/mappings/BigDecimalFieldFormatter.scala
davidfes: I agree it's not a good name, though it is existing code.

davidfes: renamed it maximumToDecimalPlaces

test/forms/mappings/MappingsSpec.scala
davidfes: fixed

test/forms/mappings/MappingsSpec.scala
davidfes: He'll check with compliance in the new year.

test/forms/mappings/MappingsSpec.scala
davidfes: All the formatters with lap have tests "must not bind values with too many decimal places"

app/forms/mappings/AdjustmentVolumesFormatter.scala
davidfes: Some formatters have the tax type where it's needed. I'm happy for ones where we know it's a specific formatter type to leave it generic like this to keep it consistent across implementations. Quite possibly when we refactor mappings this could become an inherited method rather than reimplmeent in each one. In fact I think there is far too much copy-pasta in the mappings and we could cut down the code a lot by reuse.

app/forms/mappings/BigDecimalFieldFormatter.scala
davidfes: This is an existing problem though and shouldn't block this ticket.

test/forms/mappings/MappingsSpec.scala
davidfes: Also existing problem so shouldn't block this ticket.

conf/messages.en
davidfes: We decided to keep the in litres for now for consistency with the 2dp message. Just needed the up removed.

simondrugan16 (approved):

PR #252, opened 2024-12-20
ADR-1445 found places where minus signs were spilling onto other lines and eliminated them

simondrugan16:
The ticket is https://jira.tools.tax.service.gov.uk/browse/ADR-1445. I used the nowrap css field to ensure that the minus sign and number were seen as one entity to stop the minus from going onto new lines. I created some new scala.html templates to help with this in order to avoid having to use raw html in views which i think is generally avoided on this team/on platform/scaffold in general

### DIFF
--- a/app/config/Constants.scala
+++ b/app/config/Constants.scala
@@ -67,6 +67,7 @@ object Constants {
   val visuallyHiddenCssClass     = "govuk-visually-hidden"
   val textAlignRightWrapCssClass = "text-align-right-wrap"
   val numericCellClass           = "govuk-table__cell--numeric"
+  val noWrap                     = "no-wrap"
 
   val blueTagCssClass  = "govuk-tag--blue"
   val greenTagCssClass = "govuk-tag--green"
--- a/app/config/Constants.scala
+++ b/app/config/Constants.scala
@@ -58,16 +58,14 @@ object Constants {
 
   val rowsPerPage = 15
 
-  val bodyCssClass               = "govuk-body"
-  val oneQuarterCssClass         = "govuk-!-width-one-quarter"
-  val oneHalfCssClass            = "govuk-!-width-one-half"
-  val threeQuartersCssClass      = "govuk-!-width-three-quarters"
-  val textAlignRightCssClass     = "text-align-right"
-  val boldFontCssClass           = "govuk-!-font-weight-bold"
-  val visuallyHiddenCssClass     = "govuk-visually-hidden"
-  val textAlignRightWrapCssClass = "text-align-right-wrap"
-  val numericCellClass           = "govuk-table__cell--numeric"
-  val noWrap                     = "no-wrap"
+  val bodyCssClass           = "govuk-body"
+  val oneQuarterCssClass     = "govuk-!-width-one-quarter"
+  val oneHalfCssClass        = "govuk-!-width-one-half"
+  val threeQuartersCssClass  = "govuk-!-width-three-quarters"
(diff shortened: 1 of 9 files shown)


app/config/Constants.scala
```diff
@@ -67,6 +67,7 @@ object Constants {
   val visuallyHiddenCssClass     = "govuk-visually-hidden"
   val textAlignRightWrapCssClass = "text-align-right-wrap"
   val numericCellClass           = "govuk-table__cell--numeric"
+  val noWrap                     = "no-wrap"
```
davidfes, line 70: This should be in the Css block - these constants will be deleted shortly (only 1 is in use).

app/views/components/PageHeadingWithMonetaryValue.scala.html
```diff
@@ -0,0 +1,24 @@
...
+ *@
+
+@import config.Constants.Css
+@import config.Constants
+
+@this()
+
+@(text: String, amount: String)
+
+<h1 class="@Css.headingXLCssClass">@text <span class="@Constants.noWrap">@amount</span></h1>
```
davidfes, line 24: trailing space after @text

app/views/components/ParagraphWithMonetaryValue.scala.html
```diff
@@ -0,0 +1,24 @@
...
+ *@
+
+@import config.Constants.Css
+@import config.Constants
+
+@this()
+
+@(content: Content, amount: String, classes: String = Css.bodyCssClass, id: Option[String] = None)
+
+<p class="@classes" @id.map{id => id="@id"}>@content.asHtml <span class=@Constants.noWrap>@amount</span></p>
```
davidfes, line 24: any reason the amount as bigdecimal can't be passed into here and it calls Money.format rather than rely on the caller to do the right thing?

app/views/adjustment/AdjustmentDutyDueView.scala.html
```diff
@@ -23,27 +23,30 @@
...
         sectionHeading: SectionHeading,
+        pageHeadingWithMonetaryValue: PageHeadingWithMonetaryValue,
         pageHeading: PageHeading,
         paragraph: Paragraph,
         bulletList: BulletList
 )
 
 @(adjustmentType: AdjustmentType, volume:BigDecimal, duty:BigDecimal, pureAlcoholVolume:BigDecimal, taxCode:String, rate:BigDecimal, repackagedRate:BigDecimal, repackagedDuty:BigDecimal, newDuty:BigDecimal)(implicit request: Request[_], messages: Messages)
 
+@heading = @{
```
davidfes, line 34: would this be cleaner with e.g. a dutyToShow value being created from the match, then the call to pageHeadingWithMonetaryValue just appearing once?

ajchilts (commented):
Happy once @davidfes  review comments are addressed @simondrugan16 .

app/views/components/PageHeadingWithMonetaryValue.scala.html
simondrugan16: Because this is the page title so there will be a gap between the text and the amount

app/views/components/ParagraphWithMonetaryValue.scala.html
simondrugan16: Could do. I wanted callers to have flexibility on which formatter they wanted to use here but can change it it you think it's best?

davidfes (approved):

PR #257, opened 2025-01-07
ADR-1115 increase connector test coverage with wiremock

simondrugan16:
There were a few pieces of branching logic that weren't being tested in connectors. Just a small PR to increase coverage in these areas. On Scoverage test report with ITs only, the connector coverage is slightly below 100%, but this is just because the loggers are not being printed in ITs.

### DIFF
--- a/it/test/AlcoholDutyReturnsConnectorISpec.scala
+++ b/it/test/AlcoholDutyReturnsConnectorISpec.scala
@@ -105,6 +105,32 @@ class AlcoholDutyReturnsConnectorISpec extends ISpecBase with WireMockHelper{
         }
       }
 
+      "should fail when a return is submitted, but something other than a CREATED (202) Status code is returned" in new SetUp {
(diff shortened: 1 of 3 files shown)


it/test/AlcoholDutyReturnsConnectorISpec.scala
```diff
@@ -105,6 +105,32 @@ class AlcoholDutyReturnsConnectorISpec extends ISpecBase with WireMockHelper{
         }
       }
 
+      "should fail when a return is submitted, but something other than a CREATED (202) Status code is returned" in new SetUp {
```
ajchilts, line 108: Suggest rename test to... 

"_should fail when an unexpected status code is returned_" OR;

"_submission should fail when an unexpected status code is returned_"

ajchilts (approved):
LGTM @simondrugan16  - just one suggestion.

ajchilts (approved):

PR #258, opened 2025-01-07
ADR-1648: Refactor formatters

sini-george:


ajchilts (commented):
@sini-george  @davidfes  - This change looks extensive and will require a lot of manual QA? I do not think regression  / journey tests would pick up all cases here? We may need to list out all of the affected pages and form field validators that will need to be tested to cover this off? 


@davidfes  - As you raised this ticket I feel you are best placed to review as I am new to this codebase.

app/forms/mappings/Mappings.scala
```diff
@@ -168,18 +231,119 @@ trait Mappings extends Formatters with Constraints {
...
-        maximumValueKey,
-        inconsistentKey,
+        lessOrEqualKey,
+        createVolumeFormatter(
+          requiredKey,
+          invalidKey,
+          decimalPlacesKey,
+          minimumValueKey,
+          maximumValueKey,
+          args = args
```
davidfes, line 246: no need to specify just the args naming here or in any of the others since it's not skipping parameters like some of the actual constructors.

app/forms/mappings/YearMonthFormatter.scala
```diff
@@ -95,8 +95,6 @@ class YearMonthFormatter(
   }
 
   override def unbind(key: String, value: YearMonth): Map[String, String] =
-    Map(
-      s"$key.month" -> value.getMonthValue.toString,
-      s"$key.year"  -> value.getYear.toString
-    )
+    monthIntFormatter.unbind(s"$key.month", value.getMonthValue) ++
```
davidfes, line 98: the field names month and year could go to constants,

davidfes: The bind method above is very hard to read and could do with a refactor.

davidfes: LocalDateFormatter also could have day/month/year to constants. The bind implementation there might be harder to clear up, but it probably could be a little more readable. Naming tuple values with case would help somewhat.

davidfes:
Testing should also check that existing values in the database (before the build is applied) are still loaded correctly with the field name changes in the formatters. Sinimol has already check that is the case, but it'll need signing off as part of the test as a further sanity check IMO.

app/forms/mappings/Mappings.scala
sini-george: pushed this change

app/forms/mappings/YearMonthFormatter.scala
sini-george: As discussed on chat, found LocalDateFormatter to be unused hence we can maybe delete it, and holding off making the rest of the changes until we discuss the impact on standup tomorrow.

ajchilts (changes requested):
Having gone through this ticket in terms of risk level I feel this is too risky to introduce this close to public beta. This is due to the amount of code that has changed and the amount of pages that would require manual regression prior to merging the changes. 

In addition to this the service has been pen tested and input and form validations are part of this. 

Happy to discuss at stand up tomorrow.

ajchilts:
Closing this PR due to age and number of conflicts.

PR #259, opened 2025-01-09
ADR-1380 Submission duty summary shown in entered order, not view order

simondrugan16:
[ADR-1380](https://jira.tools.tax.service.gov.uk/browse/ADR-1380) ticket was to fix the order that duty is shown on the duty due for this return page. @Bradleigh-L and I implemented this with TDD, first writing a failing test for checking the order of the items shown on this table, then implementing the suggested fix from the ticket.


ajchilts (approved):
LGTM 👍

PR #261, opened 2025-01-10
ADR-1704 Improve tests in view models

davidfes:


app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
```diff
@@ -54,20 +54,7 @@ object AdjustmentSmallProducerReliefDutyRateSummary {
             )
           )
         )
-      case (_, Some(repackagedSprDutyRate))       =>
```
davidfes, line 57: duplicates case 1

app/viewmodels/checkAnswers/adjustment/CheckYourAnswersSummaryListHelper.scala
```diff
@@ -49,10 +58,4 @@ class CheckYourAnswersSummaryListHelper @Inject() (whenDidYouPayDutySummary: Whe
         Seq(duty)
     )
   }
-
-  private def getOptionalRow(row: Option[SummaryListRow]): Seq[SummaryListRow] =
```
davidfes, line 53: reimplements .toList on Option

app/viewmodels/declareDuty/HowMuchDoYouNeedToDeclareSummary.scala
```diff
@@ -78,22 +78,21 @@ object HowMuchDoYouNeedToDeclareSummary {
...
-        createRowValues(DraughtRelief, draughtReliefRateBands, regime, dutyByTaxTypes)
+        createRowValues(draughtReliefRateBands, regime, dutyByTaxTypes)
       }
       .getOrElse(Seq.empty)
 
     coreRows ++ draughtReliefRows
   }
 
   private def createRowValues(
-    rateType: RateType,
```
davidfes, line 96: unused param

test-utils/generators/ModelGenerators.scala
```diff
@@ -224,10 +224,6 @@ trait ModelGenerators {
 
   def regimeGen: Gen[AlcoholRegime] = Gen.oneOf(AlcoholRegime.values)
 
-  def arbitraryRateBandList(regime: AlcoholRegime): Arbitrary[List[RateBand]] = Arbitrary {
```
davidfes, line 227: broken, we don't support more than 2 ranges at present - tests no longer need this.

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
```diff
@@ -25,10 +25,10 @@ import viewmodels.Money
 import viewmodels.govuk.summarylist._
 import viewmodels.implicits._
 
-object AdjustmentSmallProducerReliefDutyRateSummary {
+class AdjustmentSmallProducerReliefDutyRateSummary {
   def row(adjustmentEntry: AdjustmentEntry)(implicit messages: Messages): Option[SummaryListRow] =
     (adjustmentEntry.sprDutyRate, adjustmentEntry.repackagedSprDutyRate) match {
-      case (Some(_), Some(repackagedSprDutyRate)) =>
+      case (_, Some(repackagedSprDutyRate)) =>
```
ajchilts, line 31: Is 'Some(_)' the same as '_' ? Or is this changing the behaviour?

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
```diff
@@ -41,7 +41,7 @@ object AdjustmentSmallProducerReliefDutyRateSummary {
             )
           )
         )
-      case (Some(sprDutyRate), _)                 =>
+      case (Some(sprDutyRate), None)        =>
```
ajchilts, line 44: We are now specifically looking for 'None' here were before it was '_' (anything). Is this changing the behaviour?

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
ajchilts: This is only now a duplication of case 1 because case 1 has been modified. 

Can we be sure that the above changes are non breaking? Feels somewhat risky?

ajchilts (commented):
Just a few questions @davidfes  👍

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
davidfes: Case 1 is we have both spr and repackagedSpr
Case 2 is we have spr, but not repackagedSpr
Case 3 is we don't have spr, but we have repackaged Spr
Case 4 is we have neither

The code executed for case 1 and 3 is the same.

If we have repackagedSpr then case 1 or 3 would apply, but case 2 would not. Thus we'd land always in case 1 and 3, thus the two are identical and we don't care if we have spr or not to execute the (same) code. Thus we can remove case 3 and change the match in case 1 is ignore the value of spr (i.e. repackagedSpr takes priority).

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
davidfes: This is case 2, since we already would handle this case in case 1 if sprDutyRate was a Some(_), thus we know by this point it is None to be explicit. _ I suppose we don't care is also correct here, I just thought as None it makes it more obvious what is going on in this case.

app/viewmodels/checkAnswers/adjustment/AdjustmentSmallProducerReliefDutyRateSummary.scala
davidfes: By duplicate I mean the code executed in case 1 and 3 is the same. Please sanity check for me.

ajchilts (approved):
We got there in the end! 👍

PR #265, opened 2025-01-16
ADR-1556 Remove get/head and other issues

davidfes:


app/controllers/adjustment/CheckYourAnswersController.scala
```diff
@@ -50,13 +51,16 @@ class CheckYourAnswersController @Inject() (
...
       val result = for {
-        adjustmentEntry <- getAdjustmentEntry(request.userAnswers, index)
-        summaryList     <- checkYourAnswersSummaryListHelper.currentAdjustmentEntrySummaryList(adjustmentEntry)
-      } yield setCurrentAdjustmentEntry(request.userAnswers, adjustmentEntry, summaryList)
+        adjustmentEntry <- OptionT.fromOption[Future](getAdjustmentEntry(request.userAnswers, index))
+        summaryList     <- OptionT.fromOption[Future](
+                             checkYourAnswersSummaryListHelper.currentAdjustmentEntrySummaryList(adjustmentEntry)
+                           )
+        _               <- OptionT.liftF(setCurrentAdjustmentEntry(request.userAnswers, adjustmentEntry, summaryList))
+      } yield Ok(view(summaryList))
```
davidfes, line 59: The view was hidden in an unrelated function.

Minor side effect, set failing will go to journey recovery rather than Error Handler, but this shouldn't happen unless something is down.

app/viewmodels/tasklist/ReturnTaskListCreator.scala
```diff
@@ -45,23 +45,14 @@ class ReturnTaskListCreator @Inject() () {
     messages: Messages
   ): Section = {
     val mainTaskListItem = declareQuestionAnswer match {
-      case Some(true) =>
+      case Some(_) =>
```
davidfes, line 48: Same behaviour now, irrespective of answer to question.

app/viewmodels/tasklist/ReturnTaskListCreator.scala
```diff
@@ -219,7 +210,7 @@ class ReturnTaskListCreator @Inject() () {
     val title  = TaskListItemTitle(content = Text(messages(s"taskList.section.adjustment.$adjustmentType")))
     val status = userAnswers
       .get(AdjustmentListPage)
-      .filter(_ == false)
+      .filter(!_)
```
davidfes, line 213: sadly scala doesn't let you do .filter(_) for _ == true

app/views/checkAndSubmit/ReturnSubmittedView.scala.html
```diff
@@ -34,7 +34,7 @@
 @(viewModel: ReturnSubmittedViewModel)(implicit request: Request[_], messages: Messages)
 
 @posPanelContent = {
-    @paragraph(messages("returnSubmitted.panel.body.reference", viewModel.returnDetails.chargeReference.get.value), classes = "")
+    @paragraph(messages("returnSubmitted.panel.body.reference", viewModel.returnDetails.chargeReference.map(_.value).getOrElse("")), classes = "")
```
davidfes, line 37: or we couldn't throw because we should get a charge reference, but then the user wouldn't see the submission result.

app/controllers/adjustment/CheckYourAnswersController.scala
ajchilts: What would be the desired outcome if something is down? We should code for transient errors and as such we need to ensure we handle services being down in the correct way.  If the preferred outcome is that the user goes to journey recovery for this scenario then that's fine.

app/controllers/adjustment/CheckYourAnswersController.scala
```diff
@@ -102,14 +106,9 @@ class CheckYourAnswersController @Inject() (
...
     for {
       updateUserAnswers <- Future.fromTry(userAnswers.set(CurrentAdjustmentEntryPage, adjustmentEntry))
       _                 <- userAnswersConnector.set(updateUserAnswers)
-    } yield Ok(
-      view(
-        summaryList
-      )
-    )
-
+    } yield ()
```
ajchilts, line 113: Yielding empty parenthesis reads strangely. Although I can see why we are doing this given the OptionT refactor.

app/controllers/spiritsQuestions/SpiritTypeController.scala
```diff
@@ -90,7 +90,16 @@ class SpiritTypeController @Inject() (
     if (isNowSelected(oldValue, newValue)) {
       (answers, true)
     } else if (isNowUnselected(oldValue, newValue)) {
-      (answers.remove(OtherSpiritsProducedPage).get, false)
+      (
+        answers
+          .remove(OtherSpiritsProducedPage)
+          .getOrElse(
+            throw new RuntimeException(
+              s"Couldn't successfully remove the OtherSpiritsProduced page ${answers.returnId.appaId}/${answers.returnId.periodKey}"
```
ajchilts, line 98: This is good as leaving stale data could lead to an incorrect submission. Nice . 👍

ajchilts (approved):

app/pages/declareDuty/package.scala
```diff
@@ -21,7 +21,7 @@ import models.AlcoholRegime
 package object declareDuty {
 
   def nextPages(currentPage: QuestionPage[_]): Seq[_ <: QuestionPage[Map[AlcoholRegime, _]]] =
-    sectionPages.dropWhile(_ != currentPage).tail
+    sectionPages.dropWhile(_ != currentPage).drop(1)
```
simondrugan16, line 24: Would we want this to fail if the Seq was empty? Would it indicate a bigger issue or is it fine to do a 1 to 1 replacement here?

app/services/checkAndSubmit/AdrReturnSubmissionService.scala
```diff
@@ -374,8 +374,22 @@ class AdrReturnSubmissionServiceImpl @Inject() (
   ): EitherT[Future, String, Map[AdjustmentType, Seq[BigDecimal]]] =
     getValue(userAnswers, AdjustmentEntryListPage).map { adjustmentEntries =>
       adjustmentEntries.groupBy(_.adjustmentType).collect {
-        case (Some(RepackagedDraughtProducts), entries) => RepackagedDraughtProducts -> entries.map(_.newDuty.get)
-        case (Some(adjustmentType), entries)            => adjustmentType            -> entries.map(_.duty.get)
+        case (Some(RepackagedDraughtProducts), entries) =>
+          RepackagedDraughtProducts -> entries.map(
+            _.newDuty.getOrElse(
```
simondrugan16, line 379: Could we not pattern match on the options here and redirect to journey recovery if either of them are empty? Then we don't have to throw Exceptions

app/views/checkAndSubmit/ReturnSubmittedView.scala.html
simondrugan16: Could some of logic here move to the view model then do a warning log if the charge reference is missing, so that we know if there's an issue

app/controllers/adjustment/CheckYourAnswersController.scala
davidfes: There is a ticket for redirecting downstream stuff to the error handler, and ensuring 'they're in the wrong place' goes to journey recovery as a separate ticket that's on ice because of complexity.

app/pages/declareDuty/package.scala
davidfes: I think this code (method) probably needs a replacement (although I think it comes with the scaffold) there are a number of issues. I think the only way this could not have any pages is our error rather than something the user does so not too worried.

app/services/checkAndSubmit/AdrReturnSubmissionService.scala
davidfes: I think this would be more complicated entries is a Sequence so can't use a match, but could use a guard with forall, but then you need to repeat it in the next match otherwise it'd get handled in the non-repackaged (what if it had duty not new duty?).

 It probably is a candidate for go to journey recovery as it's a data issue though. I would consider this one as part of the journey recovery vs error handler ticket that's on ice.

Another place would be the 'journey so far' ticket, where we validate the data to reach the page. There is some validation already that we check all the tasks are complete on page load / submission, though complete doesn't necessarily mean all the expected data is there and valid (I did do some/all of this checking though).

PR #272, opened 2025-01-28
ADR-1710 Add welsh translation helper scripts and translation files

davidfes:


welsh-translation/welsh1.csv
```diff
@@ -0,0 +1,63 @@
+,English text,Welsh text,Comments,,,,
```
ajchilts, line 1: Do we need to check in all of the .csv and .key files if we generate them with the scripts? Checking these in could create stale content if not updated and also bloat the codebase? I think it will be enough to check in the README and scripts?

ajchilts (commented):
I will run through the instructions and provide further feedback.

welsh-translation/welsh1.csv
davidfes: We will need to for any we do manually which is why I checked in the others and you might want to combine with your manual changes rather than regenerate all these. There is a question on how new updates will come to us, whether new tabs or updates etc. If we get a new sheet we may or may not have to regenerate (depending on their instructions), but if we don't or it's more work not a new sheet, we may want to use what's there already rather than have to go through the whole process again.

It's a process question though.

WELSH_README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Note: Do not attempt to translate (temporarily or otherwise) anything that is missing using Google Translate etc, or fill with joke / unprofessional placeholders
+as there is a risk of it accidentally ending in production.
+In the messages.cy file the value NOT_FOUND will be used to indicate clearly something is missing which can easily be found in the file or counted automatically.
+
+## Creating key to message files from the csvs
+
+For each Welsh csv file:
+
+./create_translated_message_file.pl --messages ../conf/messages.en --translation ../welsh-translation/welsh1.csv --out ../welsh-translation/welsh1.keys
```
ajchilts, line 25: I wonder could we wrap this in a script and pass in the number of times to execute? 

I.e. 

./create-translations 8 

Saves having to execute the same command over and over?

WELSH_README.md
```diff
@@ -0,0 +1,81 @@
...
+
+You may want to save any output to help manual translation (what couldn't be translated).
+Append to the end of the command to redirect stdout and stderr to a log file:   > welsh1.log 2>&1
+
+## Create manual translations
+
+First time skip this step and came back.
```
ajchilts, line 32: Not sure what this means when I am doing this activity?

ajchilts: I assume you mean skip this step and come back after doing all of the automated work? 

If so can we move this section to some point after `## Create the message file`

welsh-translation/welsh1.csv
ajchilts: I still think as we can generate these files we do not want to check them in to Github. I think we can create them and just keep them in a directory that is ignored by Git. I feel as the spreadsheet is the source of truth we should treat it as such; making amendments there if required before repeating the process. I will have a think tomorrow about what an amendment to a specific section might look like in conjunction with this exercise.

WELSH_README.md
```diff
@@ -0,0 +1,81 @@
+# Welsh translation process
+
+This is instructions for translating the English messages file (../conf/messages.en) to the Welsh messages file (../conf/messages.cy).
+All file paths will be relative to the scripts directory where the discussed scripts are located.
+
+To run the scripts you'll need perl installed (and the path to be /usr/bin/perl) as well as the Text::CSV library/module.
+
```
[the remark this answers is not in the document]
davidfes: I didn't want to go down that route. I could point to a page (I'll see if there is anything good out there), but there might be issues for people with locked down laptops and having to have instructions for Mac, Windows, multiple Linux flavours, self-install....

WELSH_README.md
```diff
@@ -0,0 +1,81 @@
...
+## Welsh translation spreadsheet
+
+This is provided by the Welsh translation team and is 'all' the screen messages in both English and Welsh.
+This was supplied to the team in English and the Welsh translations added by them.
+
+This will be saved to the shared drive.
+On opening the spreadsheet, each tab should be saved to csv format in the ../welsh-translation directory and should be called welsh1.csv, welsh2.csv...
+where the number is the index (from 1) of the sheet from left to right.
+
```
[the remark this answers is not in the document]
davidfes: Can do, I thought users would be able to work that out.

WELSH_README.md
```diff
@@ -0,0 +1,81 @@
...
+Note: Do not attempt to translate (temporarily or otherwise) anything that is missing using Google Translate etc, or fill with joke / unprofessional placeholders
+as there is a risk of it accidentally ending in production.
+In the messages.cy file the value NOT_FOUND will be used to indicate clearly something is missing which can easily be found in the file or counted automatically.
+
+## Creating key to message files from the csvs
+
+For each Welsh csv file:
+
```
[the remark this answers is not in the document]
davidfes: I could, but on the other hand, if they can't get into the alcohol-duty-returns-frontend directory which could be anywhere on their machine. I could give an example, but I feel if you need to spell it out to this level, trying to work with perl or anything on the command line will be well above them.

WELSH_README.md
davidfes: I could write a simple shell script to do this. Let me have a think.

WELSH_README.md
davidfes: It would be the next step to do if you've done everything once already. I could put it at the end, but then you'd have to have a finish step if you've already done the next one for one before it. Perhaps a better explanation would help here.

WELSH_README.md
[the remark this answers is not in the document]
ajchilts: Can probs just ignore this comment in that regard.

hmrc-web-operations:
Thank you for submitting a pull request on this HMRC repository. A member of the HMRC team will review your changes and action accordingly.

HMRC reviewer: please reference [this runbook](https://github.com/hmrc/build-and-deploy/blob/main/runbooks/reviewing_external_pull_requests.md) for guidance on how to proceed.

PR #277, opened 2025-01-30
ADR-1687 Adding getSubscription API calls when user returns to service

matthew-goh:
Changes to logic in BeforeStartReturnController: On page load, if there are existing user answers, call the subscription and obligation APIs to check that the statuses are still valid and alcohol regimes match those in user answers. If regimes do not match, clear user answers and lock and redirect to 'service updated'.
Created new 'Service Updated' view; clicking continue creates user answers and redirects to task list.
Created a test only method to create existing user answers with the specified alcohol regimes.


app/controllers/ServiceUpdatedController.scala
```diff
@@ -0,0 +1,71 @@
...
+)(implicit ec: ExecutionContext)
+    extends FrontendBaseController
+    with I18nSupport
+    with Logging {
+
+  def onPageLoad: Action[AnyContent] = (identify andThen getData) { implicit request =>
+    Ok(view()).withSession(request.session)
+  }
+
+  // same as BeforeStartReturnController
```
ajchilts, line 50: Do we need this comment?

app/testonly/controllers/TestOnlyController.scala
```diff
@@ -16,21 +16,61 @@
...
 )(implicit val ec: ExecutionContext)
-    extends FrontendBaseController {
+    extends FrontendBaseController
+    with Logging {
 
   def clearAllData(): Action[AnyContent] = Action.async { implicit request =>
     testOnlyConnector.clearAllData().map(httpResponse => Ok(httpResponse.body))
   }
 
+  def createUserAnswers(
```
ajchilts, line 44: This is a cool feature and will be very useful when testing and creating UI journey tests 👍

app/utils/UserAnswersAuditHelper.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import com.google.inject.{Inject, Singleton}
+import models.audit.{AuditContinueReturn, AuditObligationData, AuditReturnStarted}
+import models.{ObligationData, UserAnswers}
+import play.api.Logging
+import services.AuditService
+import uk.gov.hmrc.http.HeaderCarrier
+
+import java.time.{Clock, Instant}
+
```
ajchilts, line 27: Are these new audit events? If so we will need to add them to the CIP document in confluence and seek approval.

simondrugan16: These are the audit events that we refactored out from BeforeStartReturnController.scala since they are used in the new controller too, and we didn't want to repeat ourselves. The audit events themselves are identical, but I guess they would be triggered on the Service Updated page too. I'm not sure if we need to change anything on confluence for this?

app/testonly/controllers/TestOnlyController.scala
simondrugan16: This was the only way we could think of this being testable in UI tests 😅

app/controllers/ServiceUpdatedController.scala
```diff
@@ -0,0 +1,71 @@
...
+  def onSubmit(): Action[AnyContent] = (identify andThen getData).async { implicit request =>
+    request.session.get(periodKeySessionKey) match {
+      case None            =>
+        logger.warn("Period key not present in session")
+        Future.successful(Redirect(controllers.routes.JourneyRecoveryController.onPageLoad()))
+      case Some(periodKey) =>
+        val returnAndUserDetails =
+          ReturnAndUserDetails(ReturnId(request.appaId, periodKey), request.groupId, request.userId)
+        userAnswersConnector.createUserAnswers(returnAndUserDetails).map {
```
ajchilts, line 59: @matthew-goh  @simondrugan16 - We may need to rethink the logic here 🤔 . I can envisage a scenario where the regimes do not match and we have cleared the cache and sent the user to the Service Updated page. The user then clicks the Continue button and is redirected to the task list. All good so far. The user then however (for whatever reason) hits the back button. The user will arrive back the Service Updated page. If the user then once again hits the Continue button they will get the Journey Recover page based on this logic. They will then be stuck between those two pages. 🤔 

As this page is a unique journey can we tray and get the user answers first and only create if they do not exist? If they already exist then do not create and go to task list? 

This of course makes the assumption that the Delete worked OK and that we are good to proceed with the existing answers. Unless of course we do yet another schemes comparison and proceed if they match or Journey recovery if they do not. Journey recover as our rescue mission has failed at that point and we likely do not want an endless execution loop. Say if there are issues at Mongo level etc. 

Happy to discuss. I can also demo the problem if required.

ajchilts (changes requested):

app/utils/UserAnswersAuditHelper.scala
ajchilts: I will have a think. If they are not new audit events then we should be ok. However, one will differ slightly as the url stored will be different. Let's take that up with Sharon and DA Brad.

app/controllers/ServiceUpdatedController.scala
matthew-goh: One issue with deleting in ServiceUpdatedController onPageLoad might be that if the user types service-updated in the URL, their answers will be deleted even if nothing is wrong?
So maybe onPageLoad should check if the user answers exist and redirect to journey recovery if they do, which would also cover the case where the delete somehow didn't work before redirecting to service updated?

simondrugan16: My thinking was that the service updated page already says that they need to start again. Someone going into the service to a point where they've entered answers, then pressing back to the Service Updated page would be quite a big edge case, and even then we would still be telling the user the correct info 🤷

matthew-goh: True, it would also be rare for a user to put service-updated in the URL manually after entering answers, so if we don't mind clearing their answers if that happens, I think doing DELETE onPageLoad is a good solution

ajchilts: Yes. I agree. On reflection I think whatever we do we need to make the delete and the create transactional such that if we click back from the Task List and continue again on the Service Updated the cache is cleared and created in the same operation. That we we do not get stuck at journey recovery. 

@matthew-goh  @simondrugan16  - Please reach out if you need some guidance however, I think we are agreed?

matthew-goh: Ok, we'll delete the existing answers on page load for service updated. Would we also create the new answers on page load instead of on submit? So if the user closes the page on service updated, the next time they log in, they should go to the task list without seeing the Before you start page?

matthew-goh: Or we could do the delete and create both on submit? So if the user closes the page on service updated, their old answers won't have been deleted yet and they'll see the service updated page again the next time they log in. 
This also means that if the user accidentally clicks back to service updated, if they click the browser forward button, they can return to the task list without the new answers being cleared

ajchilts: The latter sounds good to me @matthew-goh  and @simondrugan16 . Happy to retest when you get the change pushed? 👍

app/services/BeforeStartReturnService.scala
```diff
@@ -0,0 +1,57 @@
...
+    val periodKey                                                                                = userAnswers.returnId.periodKey
+    val subscriptionAndObligation: EitherT[Future, String, (Set[AlcoholRegime], ObligationData)] = for {
+      subscriptionRegimes <- alcoholDutyReturnsConnector.getValidSubscriptionRegimes(appaId)
+      obligationData      <- alcoholDutyReturnsConnector.getOpenObligation(appaId, periodKey)
+    } yield (subscriptionRegimes, obligationData)
+
+    subscriptionAndObligation.value.map {
+      case Right((subscriptionRegimes, _)) =>
+        if (userAnswers.regimes.regimes equals subscriptionRegimes) {
+          Right()
```
ajchilts, line 48: @matthew-goh  - Using `Right()` here (and in the specs) gives a compiler deprecation waring. Minor one but it would be better if we could return the positive response in a way that does not give us a warning? 

If you run `set clean test:compile` you will see the warnings.

ajchilts (changes requested):
One last comment @matthew-goh  @simondrugan16  then we are golden. I have tested the latest change and it's looking much better. Thanks.

ajchilts (dismissed):
LGTM 👍 

Great work @matthew-goh 👏 

Happy for the ticket to me moved over to test.

simondrugan16 (approved):

simondrugan16 (approved):

PR #282, opened 2025-02-04
ADR-1785 make and in 'between x% and %y' dynamic content and add Welsh helper

davidfes:


app/utils/WelshHelper.scala
```diff
@@ -0,0 +1,38 @@
...
+package utils
+
+object WelshHelper {
+
+  private def beginsWithConsonant(decimal: BigDecimal): Boolean = {
+    val wholeStr  = decimal.toBigInt.toString
+    val firstChar = wholeStr.headOption
+
+    wholeStr.length match {
+      case 1 if firstChar.contains('1') || firstChar.contains('8') => false
```
ajchilts, line 26: Do we want a function `isVowel` or something to help coders understand what we are doing here?

ajchilts (approved):
LGTM 👍

PR #289, opened 2025-02-10
ADR-1808 | Accessibility: Use ordered list with nested unordered lists on "Calculate the volume of duty suspended finished alcoholic products" page

matthew-goh:
Use an ordered list (instead of paragraphs) with nested unordered lists on "Calculate the volume of duty suspended finished alcoholic products" page - created a new view component


davidfes:
Would it be possible to rename the PR to what the changes were - the audit title won't be much use and it's on the jira.

davidfes (approved):

PR #294, opened 2025-02-11
ADR-1809 Change tables to summary lists for accessibility

matthew-goh:
Change tables to summary lists on check return page
Change Total table (previously only header) on View Return to summary list

### DIFF
--- a/app/controllers/returns/ViewReturnController.scala
+++ b/app/controllers/returns/ViewReturnController.scala
@@ -82,7 +82,7 @@ class ViewReturnController @Inject() (
             viewModel.createAlcoholDeclaredViewModel(returnDetails, ratePeriodsAndTaxCodesToRateBands)
           val adjustmentsViewModel   =
             viewModel.createAdjustmentsViewModel(returnDetails, ratePeriodsAndTaxCodesToRateBands)
-          val totalDueViewModel      = viewModel.createTotalDueViewModel(returnDetails)
+          val totalDueSummaryList    = viewModel.createTotalDueSummaryList(returnDetails)
           val netDutySuspension      = viewModel.createNetDutySuspensionViewModel(returnDetails)
           val spirits                = if (returnPeriod.hasQuarterlySpirits) {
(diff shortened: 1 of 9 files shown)


app/controllers/returns/ViewReturnController.scala
```diff
@@ -82,7 +82,7 @@ class ViewReturnController @Inject() (
             viewModel.createAlcoholDeclaredViewModel(returnDetails, ratePeriodsAndTaxCodesToRateBands)
           val adjustmentsViewModel   =
             viewModel.createAdjustmentsViewModel(returnDetails, ratePeriodsAndTaxCodesToRateBands)
-          val totalDueViewModel      = viewModel.createTotalDueViewModel(returnDetails)
```
ajchilts, line 85: I do not think we should modify the view return view for this ticket. There is nothing on the view return page that I feel should become a summary list.

ajchilts (changes requested):

app/controllers/returns/ViewReturnController.scala
matthew-goh: Removed the changes to view return, noted that ADR-1805 will change everything necessary on that page

ajchilts (dismissed):
LGTM 👍 

@Divya-Dhall  @prasanth2609  - As this has changed the layout of the page there may be journey test changes. 🤔

app/controllers/returns/ViewReturnController.scala
simondrugan16: Hi Andy, our interpretation was that 1805 would cover the total due row under the Adjustment table. This PR for 1809 should cover the table under the Table heading Total (in the pic below) that falls more under the jurisdiction of 1809 rather than 1805. I think it'd be best to add it here since it's already been implemented in the same way as the other page fixed in this PR

image

simondrugan16 (changes requested):
pls check https://github.com/hmrc/alcohol-duty-returns-frontend/pull/294#discussion_r1953010854

simondrugan16 (approved):

PR #295, opened 2025-02-12
ADR-1805 (WCAG 2.2 A - 2. Issue ID: DAC_Tables_01)

ajchilts:
Totals tables to summary lists for Adjustments and View Return.


app/views/components/TotalsTable.scala.html
```diff
@@ -0,0 +1,34 @@
...
+ *@
+
+@import config.Constants.Css
+@import viewmodels.TotalsTableViewModel
+
+@this()
+
+@(viewModel: TotalsTableViewModel, captionText: Option[String] = None)
+
+<table class="@Css.govTableClass">
```
simondrugan16, line 24: Does this solution not run into the same issue that we are tying to fix on [ADR-1809](https://jira.tools.tax.service.gov.uk/browse/ADR-1809)? The auditors issue here was that we were using a table (they took issue with the whole component) when the information shown was not a traditional table, aka it had no table headers, it's basically just a key value pair. I think this is why they suggested using a summary list on [ADR-1805](https://jira.tools.tax.service.gov.uk/browse/ADR-1805)

ajchilts: @byaqib  verified that tables are correct on this page. The problem exists as we are not rendering the totals column as part of the main table. It's in its own table. This change ensures the screen reader knows the row is a `head` of type `row` and that the value is  in a cell and not `head` . 

I am now wondering if they will simply re-raise this as a separate issue.

app/viewmodels/TotalsSummary.scala
```diff
@@ -0,0 +1,35 @@
...
+import viewmodels.govuk.summarylist._
+import viewmodels.implicits._
+
+case class TotalsSummary(key: String, value: String)
+
+object TotalsSummary {
+  def row(key: String, value: String)(implicit messages: Messages): SummaryListRow =
+    SummaryListRowViewModel(
+      key = key,
+      value = ValueViewModel(value)
```
simondrugan16, line 31: The total value, at least on the View Past Returns page, is in bold both in staging and on the prototype. The prototype is the same for the separate table we did for the total duty due table that we implemented on 1809 (screenshot below is from the prototype).

It looks like confluence, however, has them all as [non bold](https://confluence.tools.tax.service.gov.uk/display/ADR/V1a.+View+Alcohol+Duty+Return). I'm wondering if this is something we should check with UCD since there are conflicts here between the prototype and confluence.

image

ajchilts: @byaqib   - Could you please respond to the above query? Confluence seems to contradict the prototype. Which is correct for totals? Bold vs Non Bold?

simondrugan16 (approved):

PR #298, opened 2025-02-14
ADR-1803 Declare duty errors should indicate which product they refer to

davidfes:


app/forms/declareDuty/HowMuchDoYouNeedToDeclareFormProvider.scala
```diff
@@ -35,7 +35,7 @@ class HowMuchDoYouNeedToDeclareFormProvider @Inject() extends Mappings {
         "return.journey.error.minimumValue",
         "return.journey.error.maximumValue",
         "return.journey.error.lessThanExpected",
-        Seq(messages(s"return.regime.$regime"))
+        messages(s"return.regime.$regime")
```
davidfes, line 38: most specific type should be passed, with most specific name.

app/forms/mappings/BigDecimalFieldFormatter.scala
```diff
@@ -46,26 +47,34 @@ class BigDecimalFieldFormatter(
...
   private val baseFormatter = stringFormatter(s"$requiredKey.$fieldKey", args)
 
-  override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], BigDecimal] =
+  private def getKeyPrefix(key: String) = key.dropRight(fieldKey.length + 1)
+
+  override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], BigDecimal] = {
+    val keyPrefix          = getKeyPrefix(key)
+    val maybeRateBandRecap = data.get(s"$keyPrefix.$rateBandRecapField")
+
+    val allArgs = maybeRateBandRecap.fold(args)(_ +: args)
```
davidfes, line 56: Allows us to multiplex with adjustments which don't have the ratebandrecap (beer between x% and y% (311)) on them.

app/forms/mappings/VolumesAndRateFormatter.scala
```diff
@@ -111,28 +118,34 @@ class VolumesAndRateFormatter(
...
+    formatter: Formatter[T],
+    rateBandRecap: String
   ): Either[Seq[FormError], T] =
     data.get(s"$key.$field").filter(_.nonEmpty) match {
       case Some(_) => formatter.bind(s"$key.$field", data)
-      case None    => Left(Seq(requiredFieldFormError(key, field)))
+      case None    => Left(Seq(requiredFieldFormError(key, field, rateBandRecap)))
     }
 
   override def unbind(key: String, value: VolumeAndRateByTaxType): Map[String, String] =
```
davidfes, line 151: Unbind doesn't change. We don't want to change the BolumeAndRateByTaxType datastructure as it's persisted and will break any unsubmitted returns when this deploys. Luckily there is no need to have it there.

app/viewmodels/declareDuty/CategoriesByRateTypeHelper.scala
```diff
@@ -32,6 +33,10 @@ case class CategoriesByRateTypeViewModel(
 )
 
 object CategoriesByRateTypeHelper {
+  def decapitalize(s: String): String =
```
davidfes, line 36: This will disappear on the welsh capitalisation fix, so this that is dependent on this jira. Recaps are in capitalised but we don't want that when they're inserted into an error message. (Yes I know the name is Americanized, but it's a rip-off of StringOps capitalize function.

app/views/declareDuty/HowMuchDoYouNeedToDeclareView.scala.html
```diff
@@ -102,6 +102,15 @@
                     errorMessageField = form(s"volumes_${index}_taxType"),
                     value = Some(quantityViewModel.id)
                   ).hidden
+                ),
```
davidfes, line 105: This is the magic. We hide the ratebandrecap in a hidden field so the formatter can pick it up. Forms are created as schemas (volume formatters are declared as repeating) so we don't know when the formatter is created which rateband it'll store. It however can bind input fields from the html (in bind), and that's how we can get it into the formatter in order to format up the error messages. Note that the error messages are of type FormError(id, message, args*) which we cannot change and the output is not in our control either, so the ratebandrecap has to be one of the args. The second one is the regime (needed for the lpa in the beer between ... must be less than equal to total litres of beer message).

conf/messages.en
```diff
@@ -272,28 +272,28 @@ alcoholType.p1 = Select the types of alcoholic product you need to declare in th
...
-return.journey.error.maximumValue.pureAlcohol = Litres of pure alcohol must be less than 1 billion
-return.journey.error.minimumValue.pureAlcohol = Litres of pure alcohol must be more than 0.0000
-return.journey.error.tooManyDecimalPlaces.pureAlcohol = Litres of pure alcohol must be a number to 4 decimal places
-
-return.journey.error.noValue.dutyRate = Enter your Small Producer Relief duty rate
-return.journey.error.invalid.dutyRate = Your Small Producer Relief duty rate must be a number
-return.journey.error.maximumValue.dutyRate = Your Small Producer Relief duty rate must be less than £1 billion
-return.journey.error.minimumValue.dutyRate = Your Small Producer Relief duty rate must be £0.00 or more
-return.journey.error.tooManyDecimalPlaces.dutyRate = Your Small Producer Relief duty rate must be to 2 decimal places
+return.journey.error.invalid.totalLitres = Total litres of {0} must be a number
```
davidfes, line 281: confluence has this message as Total litres.

test-utils/common/TestData.scala
```diff
@@ -1124,6 +1124,8 @@ trait TestData extends ModelGenerators {
 
   val allRateBands: Set[RateBand] = allNonSmallProducerReliefRateBands ++ allSmallProducerReliefRateBands
 
+  val rateBandRecap = "recap"
```
davidfes, line 1127: could maybe use a better string for the tests but it is uniquely identifiable if needed.

test/forms/declareDuty/TellUsAboutMultipleSPRRateFormProviderSpec.scala
```diff
@@ -41,9 +41,9 @@ class TellUsAboutMultipleSPRRateFormProviderSpec extends StringFieldBehaviours w
       volumeAndRateByTaxTypes.foreach { volumeAndRateByTaxType: VolumeAndRateByTaxType =>
         val result = form.bind(
           Map(
+            "volumesWithRate.taxType"     -> volumeAndRateByTaxType.taxType,
```
davidfes, line 44: keep orders of things consistent.

test/forms/declareDuty/TellUsAboutSingleSPRRateFormProviderSpec.scala
```diff
@@ -68,294 +68,295 @@ class TellUsAboutSingleSPRRateFormProviderSpec extends StringFieldBehaviours {
...
     "must fail when empty data is provided" in {
-      val values: Map[String, String] = volumeAndRateByTaxTypes.foldRight(Map[String, String]()) {
-        (volumeAndRateByTaxType, acc: Map[String, String]) =>
-          acc ++ Map(
-            s"volumesWithRate[${volumeAndRateByTaxTypes.indexOf(volumeAndRateByTaxType)}].taxType"     -> "",
-            s"volumesWithRate[${volumeAndRateByTaxTypes.indexOf(volumeAndRateByTaxType)}].totalLitres" -> "",
-            s"volumesWithRate[${volumeAndRateByTaxTypes.indexOf(volumeAndRateByTaxType)}].pureAlcohol" -> "",
-            s"volumesWithRate[${volumeAndRateByTaxTypes.indexOf(volumeAndRateByTaxType)}].dutyRate"    -> ""
+      val values = {
+        for (index <- 0 until volumeAndRateByTaxTypes.length)
```
davidfes, line 76: This is a lot cleaner than using a fold here. Alternative is to create a Range and map for the indexes. Certainly indexOf to find the index is clunky and not a good fit for the functional paradigm.

test/forms/mappings/MappingsSpec.scala
```diff
@@ -322,6 +324,29 @@ class MappingsSpec extends SpecBase with Mappings {
       result.get mustBe AdjustmentVolumeWithSPR(BigDecimal(1234.45), BigDecimal(12.3456), BigDecimal(5.31))
     }
 
+    "must not bind when missing" - {
```
davidfes, line 327: Checks when missing are different to empty - something I missed when I wrote the tests here before.

app/forms/mappings/VolumesAndRateFormatter.scala
```diff
@@ -111,28 +118,34 @@ class VolumesAndRateFormatter(
...
   override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], VolumeAndRateByTaxType] = {
-    val taxTypeResult     = validateField(taxTypeField, key, data, taxTypeFormatter)
-    val totalLitresResult = validateField(totalLitresField, key, data, volumeFormatter)
-    val pureAlcoholResult = validateField(pureAlcoholField, key, data, pureAlcoholVolumeFormatter)
-    val dutyRateResult    = validateField(dutyRateField, key, data, dutyRateFormatter)
+    val rateBandRecap     = data
+      .getOrElse(
+        s"$key.$rateBandRecapField",
+        regimeName
+      ) // This to support MultipleSPR for which we don't know the rateBand on view creation
```
ajchilts, line 125: 👍

app/views/declareDuty/HowMuchDoYouNeedToDeclareView.scala.html
```diff
@@ -151,6 +160,15 @@
...
+                govukInput(
+                  InputViewModel(
+                    field = form(s"volumes[${index + howMuchDoYouNeedToDeclareHelper.core.size}]"),
+                    key = "rateBandRecap",
+                    label = LabelViewModel(Text("")),
+                    errorMessageField = form(s"volumes_${index + howMuchDoYouNeedToDeclareHelper.core.size}_rateBandRecap"),
+                    value = Some(quantityViewModel.rateBandRecap)
+                  ).hidden
```
ajchilts, line 171: I assume we intended to add two new inputs on this page? The do not seem to be replacing anything in the diff on the PR 🤔

ajchilts (approved):
@davidfes  -  few comments but looks good! Nice job 👍

app/forms/mappings/VolumesAndRateFormatter.scala
davidfes: yes. I already have done that, but I would want a QA to validate it.

app/views/declareDuty/HowMuchDoYouNeedToDeclareView.scala.html
davidfes: it how it gets the ratebandrecap to the formatter (see the comment above). It's hidden so the user doesn't need do anything. We do the same thing with the tax type. The html that is generated has it as an input field, but as it's hidden there is no physical box to view or change it.

PR #301, opened 2025-02-18
ADR-1787 Dyanamic rateband text and regimes shouldn't be capitalised, callers should do this as needed.

davidfes:


app/controllers/adjustment/AdjustmentDutyDueController.scala
```diff
@@ -59,9 +59,7 @@ class AdjustmentDutyDueController @Inject() (
 
   private def getView(adjustmentEntry: AdjustmentEntry)(implicit request: Request[_]): Result = {
     val result = for {
-      volume            <- adjustmentEntry.totalLitresVolume
```
davidfes, line 62: These are unneeded. We shouldn't be using them for checking if we've got them.

app/forms/declareDuty/HowMuchDoYouNeedToDeclareFormProvider.scala
```diff
@@ -35,8 +35,8 @@ class HowMuchDoYouNeedToDeclareFormProvider @Inject() extends Mappings {
         "return.journey.error.minimumValue",
         "return.journey.error.maximumValue",
         "return.journey.error.lessThanExpected",
-        messages(s"return.regime.$regime")
```
davidfes, line 38: technically we should be using regime.entryName and not relying on toString, but both work.

app/viewmodels/declareDuty/RateBandHelper.scala
```diff
@@ -54,7 +54,7 @@ object RateBandHelper {
           messages(s"return.journey.abv.interval.label.${interval.alcoholType}"),
           interval.minABV.value,
           taxType
-        ).capitalize
```
davidfes, line 57: users of this should capitalize, not this itself.

app/viewmodels/declareDuty/TellUsAboutMultipleSPRRateHelper.scala
```diff
@@ -26,12 +26,12 @@ object TellUsAboutMultipleSPRRateHelper {
     val categoryViewModels                = CategoriesByRateTypeHelper.rateBandCategories(rateBands, regime, isRecap = true)
     val smallProducerRadioItems           = categoryViewModels.smallProducer
       .map { category =>
-        RadioItem(content = Text(category.category), value = Some(category.id))
```
davidfes, line 29: .category is a bad name here

conf/messages.en
```diff
@@ -262,11 +262,11 @@ beforeStartReturn.text.dueDateWarning = The deadline for completing this return
 
 alcoholType.title = Which types of alcoholic products do you need to declare?
 alcoholType.heading = Which types of alcoholic products do you need to declare?
-alcoholType.Beer = Beer
```
davidfes, line 265: these should be capitalised as necessary

conf/messages.en
```diff
@@ -306,18 +306,18 @@ return.journey.abv.interval.exceeding.max = {0} at or above {1}% ABV (tax type c
 return.journey.abv.single.interval = {0} between {1}% {2} {3}% ABV (tax type code {4})
 return.journey.abv.multi.interval = {0} between {1}% {2} {3}% ABV and {4} between {5}% {6} {7}% ABV (tax type code {8})
 
-return.journey.abv.recap.interval.exceeding.max.Core = Non-draught {0} at or above {1}% ABV ({2})
```
davidfes, line 309: same here

test/controllers/adjustment/SpoiltAlcoholicProductTypeHelperSpec.scala
```diff
@@ -61,10 +61,10 @@ class SpoiltAlcoholicProductTypeHelperSpec extends SpecBase {
           spoiltAlcoholicProductTypeHelper.radioOptions(AlcoholRegimes(Set(Cider, OtherFermentedProduct, Beer)))
 
         radioOptions mustBe Seq(
-          RadioItem(Text(messages("alcoholType.Beer")), Some("Beer"), Some("Beer")),
```
davidfes, line 64: IMO tests shouldn't be looking up messages.

test/controllers/adjustment/AdjustmentDutyDueControllerSpec.scala
```diff
@@ -185,9 +185,7 @@ class AdjustmentDutyDueControllerSpec extends SpecBase {
     }
 
     val incompleteAdjustmentEntries = List(
-      (adjustmentEntry.copy(totalLitresVolume = None), "totalLitres"),
```
davidfes, line 188: these were unused thus removed.

conf/messages.en
davidfes: since we can safely move these, renamed to regime.<regime>  ofp has to be lowercase due to toLowerCase() call

app/viewmodels/declareDuty/TellUsAboutMultipleSPRRateHelper.scala
davidfes: we should, but the problem is 'shaving a yak' i.e. it unrolls with all the things you find (then you need to be sure you didn't break something). Happy to do a small tech refactor for things like this but outside of this change.

test/controllers/adjustment/AdjustmentDutyDueControllerSpec.scala
davidfes: it was just accessing them unnecessarily (probably part of cut and paste). The most 'benefit' was checking they were there but they shouldn't be needed and if there is a need to check for completeness (which we do elsewhere) it should be a specific check.

matthew-goh (dismissed):

ajchilts (approved):

PR #305, opened 2025-02-21
ADR-1786 Support 'of <regime>' soft mutations in Welsh

davidfes:


app/forms/declareDuty/HowMuchDoYouNeedToDeclareFormProvider.scala
```diff
@@ -37,6 +37,6 @@ class HowMuchDoYouNeedToDeclareFormProvider @Inject() extends Mappings {
         "return.journey.error.lessThanExpected",
         messages(regime.regimeMessageKey)
```
matthew-goh, line 38: In the error messages where regime name is passed in directly (in this case return.journey.error.lessThanExpected), would we have to use regimeMessageSoftMutationKey here?

app/forms/declareDuty/TellUsAboutSingleSPRRateFormProvider.scala
```diff
@@ -37,6 +37,6 @@ class TellUsAboutSingleSPRRateFormProvider @Inject() extends Mappings {
         "return.journey.error.lessThanExpected",
         messages(regime.regimeMessageKey)
```
matthew-goh, line 38: For the error messages where regime name is passed in directly (in this case return.journey.error.lessThanExpected), would we have to use regimeMessageSoftMutationKey here?

matthew-goh (commented):
For the multiple SPR form, regime name comes after 'of' in the total litres error messages but not the pure alcohol ones. So the form formatter probably needs edits to accommodate this...

matthew-goh (approved):

PR #306, opened 2025-02-21
ADR-1834 Error summary link to checkbox and inline error message on "What do you need to declare?" page

matthew-goh:


app/views/declareDuty/WhatDoYouNeedToDeclareView.scala.html
```diff
@@ -38,7 +39,7 @@
     @formHelper(action = controllers.declareDuty.routes.WhatDoYouNeedToDeclareController.onSubmit(mode, regime), Symbol("autoComplete") -> "off") {
 
         @if(form.errors.nonEmpty) {
-            @govukErrorSummary(ErrorSummaryViewModel(form, errorLinkOverrides = Map("value" -> "value_0")))
+            @govukErrorSummary(ErrorSummaryViewModel(form, errorLinkOverrides = Map("rateBand" -> taxBandsViewModel.core.headOption.flatMap(_.id).getOrElse(""))))
```
matthew-goh, line 42: This links to the first checkbox, which in this case is always the first rate band under taxBandsViewModel.core.
Similar to other forms where at least one checkbox must be selected (e.g. Which types of alcoholic products do you need to declare?)

app/views/declareDuty/WhatDoYouNeedToDeclareView.scala.html
```diff
@@ -51,6 +52,13 @@
 
                 @paragraph(messages("whatDoYouNeedToDeclare.hint"), classes = Css.hintCssClass)
 
+                @if(form.errors.nonEmpty) {
+                    @errorMessage(
+                        errorText = messages(form.errors.head.messages.headOption.getOrElse("")),
+                        id = Some(s"${form.errors.head.key}-error")
+                    )
+                }
+
```
matthew-goh, line 61: Since the error message does not relate to a specific input field or group of checkboxes (there are 4 groups of checkboxes on the page), I created a new component for a standalone inline error message, which is displayed here if the error has occurred.

app/views/components/ErrorMessageStandalone.scala.html
```diff
@@ -0,0 +1,26 @@
...
+@import config.Constants.Css
+
+@this()
+
+@(errorText: String, id: Option[String] = None)(implicit messages: Messages)
+
+<p @id.map{id => id="@id"} class="@Css.errorMessageCssClass">
+    <span class="@Css.visuallyHiddenCssClass">@messages("error.prefix"):</span>
+    @errorText
+</p>
```
matthew-goh, line 26: Following HTML on https://design-system.service.gov.uk/components/error-message/

ajchilts:
This looks to solve the problem so am happy to approve. 👍 
Screenshot 2025-02-24 at 13 27 05
Screenshot 2025-02-24 at 13 26 55

ajchilts (approved):
LGTM - tested locally and attached screenshots.

app/views/declareDuty/WhatDoYouNeedToDeclareView.scala.html
simondrugan16, line 42: What would happen if this id was empty?

app/views/components/ErrorMessageStandalone.scala.html
simondrugan16: should there be a space after the colon in the message? Not sure if this would effect the way this gets read out by a screenreader ("Error:Error" vs "Error: Error")

app/views/declareDuty/WhatDoYouNeedToDeclareView.scala.html
matthew-goh: We would have Map("rateBand" -> "") and the link wouldn't go to any checkbox. Other forms with checkboxes do it similarly, e.g. Which types of alcoholic products do you need to declare?, Which of these spirits have you produced?

PR #307, opened 2025-02-21
ADR-1836: Fix hidden text for Usability and edit test.

Bradleigh-L:


matthew-goh (commented):
The new accessibility report mentions complete-return/quarterly-spirits-production/check-your-answers as another instance of this issue - the visually hidden text there makes sense, but doesn't match the summary list keys exactly. Should we change that too?

conf/messages.en
```diff
@@ -642,7 +642,7 @@ declareSpiritsTotal.para2.list2 = do not include feints remaining at the end of
 declareSpiritsTotal.checkYourAnswersLabel = Total of all spirits
 declareSpiritsTotal.error.nonNumeric = Total volume of spirits must be a number
 declareSpiritsTotal.error.required = Enter the total volume of spirits taken in litres of pure alcohol
-declareSpiritsTotal.change.hidden = the total volume of spirits you made this quarter
+declareSpiritsTotal.change.hidden =  total of all spirits
```
matthew-goh, line 645: Small thing but there's an extra space after the = !
I think we'd also change otherSpiritsProduced.change.hidden to match the key 'Other spirits produced'

matthew-goh (approved):
Approved after confirming in dev catch-up - since the spirits table was mentioned, they must want us to fix it like this

PR #308, opened 2025-02-21
ADR-1853 Support dynamic error messages to be translated to Welsh

davidfes:


matthew-goh (approved):
Seems good, also when running the frontend :)

ajchilts (changes requested):
Lets hold off merging anything else @davidfes until we have discussed the best foot forward given the discussion with the centre tech leads on 26/02.

davidfes:
Can close this - we're using the message file to do the work not the code.

PR #310, opened 2025-02-26
ADR-1509 Ensure all required info is populated when spirits check your answers page loads

matthew-goh:
Redirect to journey recovery if no spirit type is selected or if "Other spirits" is selected but no details are provided

### DIFF
--- a/app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
+++ b/app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
@@ -21,7 +21,7 @@ import models.UserAnswers
 import pages.spiritsQuestions.SpiritTypePage
 import play.api.i18n.Messages
 import uk.gov.hmrc.govukfrontend.views.Aliases.Text
-import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.{Card, CardTitle, SummaryList, SummaryListRow}
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.{Card, CardTitle, SummaryList}
 import viewmodels.govuk.summarylist.SummaryListViewModel
 
 object CheckYourAnswersSummaryListHelper {
--- a/app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
+++ b/app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
@@ -16,7 +16,9 @@
 
 package viewmodels.checkAnswers.spiritsQuestions
 
+import models.SpiritType.Other
 import models.UserAnswers
+import pages.spiritsQuestions.SpiritTypePage
 import play.api.i18n.Messages
 import uk.gov.hmrc.govukfrontend.views.Aliases.Text
 import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.{Card, CardTitle, SummaryList, SummaryListRow}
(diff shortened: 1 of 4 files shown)


simondrugan16:
PR is looking good. Did you check if any coverage needed to be added to CheckYourAnswersSummaryListHelperSpec.scala? I get that the controller spec is essentially testing this already (which it shouldn't, but that's the testing pattern that was chosen for the whole project) but it'd be worth double checking

matthew-goh:
There is currently no CheckYourAnswersSummaryListHelperSpec for spirits! I guess they thought it wasn't needed if it's being tested in the controller spec already. But we could create one if that would be better?

simondrugan16:
I think that'd be worth doing, yeah :)

app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
```diff
@@ -16,34 +16,37 @@
 
 package viewmodels.checkAnswers.spiritsQuestions
 
+import models.SpiritType.Other
 import models.UserAnswers
+import pages.spiritsQuestions.SpiritTypePage
 import play.api.i18n.Messages
 import uk.gov.hmrc.govukfrontend.views.Aliases.Text
 import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.{Card, CardTitle, SummaryList, SummaryListRow}
```
ajchilts, line 24: Unused import? SummaryListRow

ajchilts (changes requested):

app/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelper.scala
ajchilts: If you could please run 'sbt runAllChecks' and fix the warning @matthew-goh ? Thanks

ajchilts (commented):
LGTM 👍

ajchilts (dismissed):

test/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelperSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+        .set(
+          SpiritTypePage,
+          Set[SpiritType](SpiritType.Maltspirits, SpiritType.Grainspirits)
+        )
+        .success
+        .value
+
+      val expectedSummaryListKeys = List("Total of all spirits", "Scotch Whisky", "Irish Whiskey", "Type of spirits")
+
+      val summaryListOption = CheckYourAnswersSummaryListHelper.spiritsSummaryList(userAnswers)(getMessages(app))
```
simondrugan16, line 76: you can just do a .get here

test/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelperSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+
+      "spirit type page (always required) is not populated" in {
+        val userAnswers = completedUserAnswers
+          .remove(List(OtherSpiritsProducedPage, SpiritTypePage))
+          .success
+          .value
+
+        val summaryListOption = CheckYourAnswersSummaryListHelper.spiritsSummaryList(userAnswers)(getMessages(app))
+
+        summaryListOption mustBe None
```
simondrugan16, line 96: I think this test gives a clear example of why this helper was poorly designed in the first place. We have missing data and this method just returns none... I blame scaffold :D

test/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelperSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      "Other spirits is selected but other spirits produced page is not populated" in {
+        val userAnswers = completedUserAnswers
```
simondrugan16, line 100: userAnswers -> userAnswersWithoutOtherSpiritsProducedPage ?

test/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelperSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+    .value
+
+  "CheckYourAnswersSummaryListHelper" - {
+
+    "must return a summary list with the correct rows if pages are populated (Other spirits is selected)" in {
+      val expectedSummaryListKeys =
+        List("Total of all spirits", "Scotch Whisky", "Irish Whiskey", "Type of spirits", "Other spirits produced")
+
+      val summaryListOption =
+        CheckYourAnswersSummaryListHelper.spiritsSummaryList(completedUserAnswers)(getMessages(app))
```
simondrugan16, line 52: .get here too

test/viewmodels/checkAnswers/spiritsQuestions/CheckYourAnswersSummaryListHelperSpec.scala
matthew-goh: If we use .get and the method returns None, the test would fail with an exception. Is that ok?
Currently it allows us to examine the value inside the option and test that further, but fail without throwing an exception if it is None (and the val summaryListOption line is consistent across all tests).

simondrugan16: From what I can see in other places they just wrap the expected value in a Some rather than pattern match. Matching on the option here adds noise to the test where it is not really needed or particularly helpful. Not really a big deal though so up to you

simondrugan16 (approved):

PR #312, opened 2025-02-28
ADR-1581 Change clock to UK time zone and use injected clock where possible

matthew-goh:
Injected clock is changed in config.Module (defined constant for time zone string ID as there isn't a preexisting one)

### DIFF
--- a/app/config/Module.scala
+++ b/app/config/Module.scala
@@ -19,7 +19,7 @@ package config
 import com.google.inject.AbstractModule
 import controllers.actions._
 
-import java.time.{Clock, ZoneOffset}
+import java.time.{Clock, ZoneId}
 
 class Module extends AbstractModule {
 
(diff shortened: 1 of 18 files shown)


app/config/Module.scala
```diff
@@ -37,6 +37,6 @@ class Module extends AbstractModule {
       .asEagerSingleton()
 
     bind(classOf[ServiceEntryCheckAction]).to(classOf[ServiceEntryCheckActionImpl]).asEagerSingleton()
-    bind(classOf[Clock]).toInstance(Clock.systemDefaultZone.withZone(ZoneOffset.UTC))
+    bind(classOf[Clock]).toInstance(Clock.system(ZoneId.of(Constants.ukTimeZoneStringId)))
```
ajchilts, line 40: It will be worth testing the time logged both locally and on the server in staging once this is merged. We need to ensure we do not violate the date time as dictated by the server. I.e. make sure the clock is not out when hosted in an environment.

ajchilts (approved):

ajchilts (approved):

PR #314, opened 2025-03-04
ADR-1529: Refactor the AdjustmentNavigator

Bradleigh-L:


app/navigation/AdjustmentNavigator.scala
```diff
@@ -152,12 +154,13 @@ class AdjustmentNavigator @Inject() () {
...
-      normalRoutes(page)(userAnswers)
-    case CheckMode  =>
-      checkRouteMap(page)(userAnswers)(hasAnswerChanged)
-  }
+  def nextPage(page: Page, mode: Mode, userAnswers: UserAnswers, hasAnswerChanged: Option[Boolean] = None): Call =
+    mode match {
+      case NormalMode =>
+        normalRoutes(page)(userAnswers)
+      case CheckMode  =>
+        checkRouteMap(page)(userAnswers)(hasAnswerChanged)
```
simondrugan16, line 162: In here I would probably change `hasAnswerChanged` => `hasAnswerChanged.getOrElse(false)`, then above, change all the `hasAnswerChanged` => `needToChangeOtherAnswersToo` (that's a terrible name but I hope you catch my drift).

app/navigation/AdjustmentNavigator.scala
```diff
@@ -152,12 +154,13 @@ class AdjustmentNavigator @Inject() () {
     }
   }
 
-  def nextPage(page: Page, mode: Mode, userAnswers: UserAnswers, hasAnswerChanged: Boolean = true): Call = mode match {
-    case NormalMode =>
-      normalRoutes(page)(userAnswers)
-    case CheckMode  =>
-      checkRouteMap(page)(userAnswers)(hasAnswerChanged)
-  }
+  def nextPage(page: Page, mode: Mode, userAnswers: UserAnswers, hasAnswerChanged: Option[Boolean] = None): Call =
```
simondrugan16, line 157: I would remove the default value for `hasAnswerChanged` here and go through all the places it's used and define it explicitly. Going from defaulting to true, to defaulting to essentially false it a bit dangerous if there were places where it mattered that it was true, but just weren't defining it because it was already defaulted in

PR #316, opened 2025-03-07
ADR-1889 Prevent double click on buttons

matthew-goh:


app/viewmodels/govuk/ButtonFluency.scala
```diff
@@ -60,8 +61,8 @@ trait ButtonFluency {
     def disabled(): Button =
       button.copy(disabled = true)
 
-    def preventingDoubleClick(): Button =
-      button.copy(preventDoubleClick = Some(true))
+    def allowingDoubleClick(): Button =
+      button.copy(preventDoubleClick = None)
```
matthew-goh, line 65: Since preventDoubleClick will be enabled by default, change this method to allow double click instead

simondrugan16: maybe rename to buttonWithDoubleClick?

simondrugan16 (approved):

PR #324, opened 2025-03-17
ADR-1516 Make legend M size rather than S on multiple spr page

davidfes:


ajchilts (approved):
LGTM 👍 
@davidfes  - PR builder failed so have kicked it off again.

PR #326, opened 2025-03-26
ADR-1911 Fix flaky DutyCalculationHelperSpec test

matthew-goh:
1. Ensure that `unmatchedRateBand` has a different tax type code from those in `rateBands`
- Separate method to generate unmatchedRateBand in which `taxTypeCode` is a string of digits (instead of an alphaStr)
2. Ensure that generators don't produce bad data from which sampling gives `None`: 
- Change `suchThat(_.nonEmpty)` to `retryUntil(_.nonEmpty)` for `taxTypeCode`
- `minABV` and `maxABV` must come from a list of ranges

Test passes when run 50000 times:
ADR-1911 tests passed


davidfes (approved):

ajchilts:
Passed QA. Happy to merge.

PR #327, opened 2025-03-27
ADR-1903 | Only refresh User Answers on session timeout where there is a period key in the session

ajchilts:


test/controllers/KeepAliveControllerSpec.scala
```diff
@@ -74,5 +74,22 @@ class KeepAliveControllerSpec extends SpecBase {
         }
```
matthew-goh, line 74: Perhaps we can add a verify to the first 2 tests to assert that mockUserAnswersConnector.keepAlive was called?

matthew-goh:
Tested locally and it works :)

test/controllers/KeepAliveControllerSpec.scala
ajchilts: We could indeed.

PR #331, opened 2025-03-31
ADR-1923 Make adjustment type in hidden text non-dynamic on 'Adjustments from previous returns'

matthew-goh:
Welsh hidden text:
ADR-1923 Welsh hidden text


ajchilts (approved):
LGTM

PR #333, opened 2025-04-01
ADR-1876 Avoid programmer awareness of Welsh grammar in rateband messages

Bradleigh-L:


test-utils/common/TestData.scala
```diff
@@ -1025,7 +1025,7 @@ trait TestData extends ModelGenerators {
     Set(
       RangeDetailsByRegime(
         Beer,
-        NonEmptySeq.one(ABVRange(AlcoholType.Beer, AlcoholByVolume(BigDecimal(1)), AlcoholByVolume(BigDecimal(2))))
+        NonEmptySeq.one(ABVRange(AlcoholType.Beer, AlcoholByVolume(BigDecimal(1)), AlcoholByVolume(BigDecimal(3))))
```
ajchilts, line 1028: Assuming the change from `2` to `3` (increment) is intended here?

test-utils/common/TestData.scala
```diff
@@ -1116,7 +1116,7 @@ trait TestData extends ModelGenerators {
...
```
ajchilts, line 1119: As above for all. Seem to be incremented by 1? Intended?

test/viewmodels/checkAnswers/adjustment/AdjustmentRepackagedTaxTypeSummarySpec.scala
```diff
@@ -28,7 +28,7 @@ class AdjustmentRepackagedTaxTypeSummarySpec extends SpecBase {
       adjustmentRepackagedTaxTypeSummary.row(adjustmentEntry) mustBe Some(
         SummaryListRow(
           Key(Text("New tax type")),
-          Value(Text("Non-draught beer between 1% and 2% ABV (tax type code 123)")),
+          Value(Text("Non-draught beer between 1% and 3% ABV (tax type code 123)")),
```
ajchilts, line 31: As above? Intended increment?

ajchilts (commented):
@Bradleigh-L  @davidfes  - I am assuming the increments on the right of ABV are intended?

test-utils/common/TestData.scala
davidfes: yes, it now throws an exception if the upper band is not one of those we know about (as discussed) thus the tests also have to conform.

ajchilts (approved):
LGTM 👍

PR #337, opened 2025-04-02
ADR-1894: Update content on /no-approval-id

Bradleigh-L:


matthew-goh:
If wasReferredFromBTA=true, the "Return to your business tax account" button is currently green, but it's grey in the screenshot on the ticket. I guess we have to change it to a secondary button

app/views/auth/NoAppaIdView.scala.html
```diff
@@ -43,11 +43,13 @@
             href = appConfig.appaIdRegisterUrl,
             newTab = true
             )
+            @messages("noAppaId.p2")
         </p>
 
         @if(wasReferredFromBTA) {
             @govukButton(
                 ButtonViewModel("return-to-business-account",messages("site.returnToBusinessAccount"))
+                .withCssClass("govuk-button--secondary")
```
matthew-goh, line 52: This works but since there's a secondaryButtonCssClass in Constants, we could use that? :)

matthew-goh (dismissed):

PR #339, opened 2025-04-03
ADR-1907 Remove unused messages

davidfes:


app/forms/mappings/Formatters.scala
```diff
@@ -68,29 +68,26 @@ trait Formatters {
...
-
-      override def bind(key: String, data: Map[String, String]) =
+      override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], Boolean] =
         baseFormatter
           .bind(key, data)
           .flatMap {
             case "true"  => Right(true)
             case "false" => Right(false)
-            case _       => Left(Seq(FormError(key, invalidKey, args)))
+            case _       => throw new IllegalArgumentException(s"Excepted a boolean for $key")
```
davidfes, line 87: this can only happen with bad data, it's not a user error

app/forms/mappings/Formatters.scala
```diff
@@ -119,22 +115,22 @@ trait Formatters {
...
-      private val baseFormatter = stringFormatter(requiredKey, args)
+      private val baseFormatter = stringFormatter(requiredKey)
 
       override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], A] =
         baseFormatter.bind(key, data).flatMap { str =>
           ev.withName(str)
-            .map(Right.apply)
-            .getOrElse(Left(Seq(FormError(key, invalidKey, args))))
+            .map(Right[Seq[FormError], A])
+            .getOrElse(throw new IllegalArgumentException(s"Expected an enumeration value for $key"))
```
davidfes, line 133: again bad data, not a user error.

app/views/adjustment/WhenDidYouPayDutyView.scala.html
```diff
@@ -35,10 +35,10 @@
 @(form: Form[_], mode: Mode, adjustmentType: AdjustmentType, exciseEnquiriesUrl: String)(implicit request: Request[_], messages: Messages)
 
 @heading = @{
-    if(adjustmentType == Underdeclaration) {
+    if (adjustmentType == Underdeclaration) {
         messages("whenDidYouPayDuty.under-declaration.title")
-    } else if(adjustmentType == Spoilt) {
-        messages("whenDidYouPayDuty.spoilt.title")
+    } else if (adjustmentType == Spoilt) {
+        throw new IllegalArgumentException("Not expecting WhenDidYouPayDuty to be asked for Spoilt")
```
davidfes, line 41: shouldn't get here in the current journey since we don't ask for when with spoilt

ajchilts: Doing more than removing unused messages here? Did we used to ask this question? Is this a hangover from an old journey?

conf/app.routes
```diff
@@ -13,7 +13,6 @@ GET         /there-is-a-problem                          controllers.JourneyReco
 GET         /return-locked                               controllers.ReturnLockedController.onPageLoad()
 
 GET         /account/sign-out-survey                     controllers.auth.AuthController.signOut()
-GET         /account/signed-out                          controllers.auth.SignedOutController.onPageLoad
```
ajchilts, line 16: This is doing more than removing unused messages. Are we 100% sure this is not used?

ajchilts (commented):
A couple of questions @davidfes however, looks good after clarification.

app/views/adjustment/WhenDidYouPayDutyView.scala.html
davidfes: yes, the spoilt journey used to be different.

PR #341, opened 2025-04-08
ADR-1957 | Improve controller tests that should use mocks

Bradleigh-L:


app/viewmodels/declareDuty/CheckYourAnswersSPRSummaryListHelper.scala
```diff
@@ -1,32 +0,0 @@
...
-import models.{AlcoholRegime, UserAnswers}
-import play.api.i18n.Messages
-import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryList
-
-object CheckYourAnswersSPRSummaryListHelper {
-
-  def summaryList(regime: AlcoholRegime, userAnswers: UserAnswers, index: Option[Int])(implicit
-    messages: Messages
-  ): Option[SummaryList] = {
-    val rows = TellUsAboutMultipleSPRRateSummary.rows(regime, userAnswers, index)
```
davidfes, line 28: This really wasn't adding anything and the wrapping can be done in the controller.

ajchilts:
@Bradleigh-L 

Closing this PR as won't do as...

- Its not critical to live handover 
- Its currently based on scaffolded code
- Its on pattern as is
- We can't so all the tests
- Any new generated (scaffolded) code would need the same changes going forward
- ECP would need the same changes
- Too difficult to enforce / maintain this approach which will ultimately lead to inconsistencies in the approach and code base
- we have more pressing priorities

PR #342, opened 2025-04-09
ADR-1996 Prevent pre-population of multiple SPR form

matthew-goh:
Clear TellUsAboutMultipleSPRRatePage in MultipleSPRListController.onSubmit() in the same way as CheckYourAnswersSPRController, in case the user has arrived at the multiple SPR list page from 'Change' but without changing any values


ajchilts:
@matthew-goh  - tested locally and it is looking good. 👍

ajchilts (approved):
LGTM 👍

PR #344, opened 2025-04-09
ADR-1999 Fix error message on '<Regime> with Small Producer Relief duty to declare'

matthew-goh:
Correct English: Select yes if you have another Small Producer Relief duty rate to declare
Correct Welsh: Dewiswch ‘Iawn’ os oes gennych gyfradd doll arall yn sgil Rhyddhad i Gynhyrchwyr Bach i’w datgan


ajchilts (approved):
LGTM 👍

PR #345, opened 2025-04-10
ADR-1997 Add feature switch for new DSD journey

matthew-goh:
Includes controller action to check feature switch that will be applied on page load in the new journey (similar to the one for spirits)


matthew-goh:
Version update done in ADR-1999

ajchilts (approved):
LGTM 👍

PR #349, opened 2025-04-11
ADR 1900 Error page when user signed in as individual instead of Organisation

simondrugan16:
[ADR-1900](https://jira.tools.tax.service.gov.uk/browse/ADR-1900) DEV: New page Error onward journey issue when still signed indiv instead of Organisation

Changed the Non Enrolled action to take the user to an specific error page when the user is signed in with the wrong sort of account (aka not signed in as an Organisation) as this is a special case error. In this scenario we present them with a new error page with a button that when pressed, signs them out and allows them to sign back in with the correct account. If they do this successfully, we progress them to enrolment management frontend where they can continue their journey.


ajchilts (approved):
LGTM @simondrugan16  👍 . Nice bit of renaming too! Much easier to follow.

PR #351, opened 2025-04-14
ADR-1703 New DSD journey: Which finished alcoholic products have you delivered or received under duty suspense?

matthew-goh:
Second page in new DSD journey.

Screenshots:
ADR-1703 English

ADR-1703 Welsh


conf/messages.cy
```diff
@@ -105,7 +105,7 @@ section.adjustment.spoilt = Addasu ar gyfer alcohol wedi’i ddifetha
 section.adjustment.drawback = Addasu ar gyfer ad-dalu toll
 section.adjustment.repackaged-draught-products = Addasu ar gyfer cynhyrchion o’r gasgen wedi’u hailbecynnu
 section.adjustment.checkAnswers = Addasiad o ddatganiad blaenorol
-section.dutySuspended = Dosbarthiadau o gynhyrchion alcoholaidd gorffenedig y gohiriwyd tollau arnynt
+section.dutySuspended = Rhoi gwybod am ddosbarthiadau o alcohol y gohiriwyd tollau arno
```
matthew-goh, line 108: Welsh message was previously for the old caption “Duty suspended finished alcoholic product deliveries”

test-utils/common/TestData.scala
```diff
@@ -577,6 +578,63 @@ trait TestData extends ModelGenerators {
     )
   )
 
+  val userAnswersWithDutySuspendedData = emptyUserAnswers.copy(data =
+    Json.obj(
```
matthew-goh, line 582: This is how I imagine the data will be structured after the user has submitted answers in the new journey.
Similar to AlcoholDutyPage from declareDuty, each regime will be a key in (1) DutySuspendedQuantitiesPage where the values entered by the user will be stored, and (2) DutySuspendedFinalVolumesPage where the calculated volumes (to be submitted) will be stored.

If regimes are removed when selecting alcohol types, this allows us to use removePagesByKey to clear the data for those regimes, regardless of which regimes are removed.

ajchilts: Agreed and its test data so can be subject to change as we progress if needs be.

ajchilts (approved):
LGTM - nice wok moving over to the clean up functions too.  👍

matthew-goh:

PR #356, opened 2025-04-17
ADR-1780 New DSD journey: Declare quantity page

matthew-goh:
Page will be shown for each regime that the user has selected, although the navigation from one regime to the next has not been implemented here as the next page (Display calculation for the regime) does not exist yet

Includes the call to alcohol-duty-calculator to calculate the total litres and pure alcohol that will be submitted in the return


app/navigation/DutySuspendedNavigator.scala
```diff
@@ -20,40 +20,76 @@ import controllers._
...
-    // TODO: Implement for new pages that depend on the regime
-    _ => _ => _ => routes.JourneyRecoveryController.onPageLoad()
+  private val normalRoutesWithRegime: Page => UserAnswers => AlcoholRegime => Call = {
+    case DutySuspendedQuantitiesPage =>
+      userAnswers =>
+        regime =>
+          userAnswers.getByKey(DutySuspendedQuantitiesPage, regime) match {
+            case Some(_) =>
+              // TODO: Go to display calculation page for current regime
+              logger.info("Go to display calculation page")
```
ajchilts, line 56: I assume we will remove this log statement when we complete the implementation? If so might be better as 'logger.debug'?

ajchilts (approved):
LGTM @matthew-goh  👍

matthew-goh:

PR #357, opened 2025-04-25
ADR-1781 New DSD journey: Display calculation page

matthew-goh:


app/navigation/DutySuspendedNavigator.scala
```diff
@@ -71,22 +121,21 @@ class DutySuspendedNavigator @Inject() () {
...
+      checkRouteMap(page)(userAnswers)(hasRegimesToAdd.getOrElse(false))
   }
 
-  private def declareDutySuspenseQuestionPageRoute(userAnswers: UserAnswers, mode: Mode): Call =
+  private def declareDutySuspenseQuestionPageRoute(userAnswers: UserAnswers): Call =
     userAnswers.get(DeclareDutySuspenseQuestionPage) match {
       case Some(true) if userAnswers.regimes.regimes.size > 1 =>
-        // TODO: Go to new declare quantity page for the single regime (insert mode on page load)
-        routes.JourneyRecoveryController.onPageLoad()
+        controllers.dutySuspendedNew.routes.DutySuspendedAlcoholTypeController.onPageLoad(NormalMode)
```
matthew-goh, line 135: If the user answers 'yes' on the initial question page, they will always go through the rest of the journey in normal mode (with existing answers pre-populated if available) - this is consistent with the old DSD journey and the spirits journey. For pages after the initial question, check mode is accessed only through Check Your Answers.

app/navigation/DutySuspendedNavigator.scala
```diff
@@ -71,22 +121,21 @@ class DutySuspendedNavigator @Inject() () {
...
       case Some(true)                                         =>
-        // TODO: Go to new alcohol type selection page (insert mode on page load)
-        routes.JourneyRecoveryController.onPageLoad()
+        controllers.dutySuspendedNew.routes.DutySuspendedQuantitiesController
+          .onPageLoad(NormalMode, userAnswers.regimes.regimes.head)
```
ajchilts, line 138: Can we always guarantee we have regimes? I.e. `head` will never raise an exception?

ajchilts (approved):

app/navigation/DutySuspendedNavigator.scala
matthew-goh: There should always be regimes at this point - if the user isn't approved for any regimes, user answers can't be created as reading the subscription summary causes this exception in alcohol-duty-returns:
IllegalArgumentException("Expecting at least one regime to be approved")

matthew-goh:

PR #359, opened 2025-04-29
ADR-1788 New DSD journey: Updates to task list and return submission

matthew-goh:
- Updated the task list logic to check pages from the new journey if the feature switch is on. If the user has filled in the old journey but not submitted, the DSD section shows as incomplete.
- Updated the "Duty due" page logic, as well as the creation of the return, to check/read the new pages
- Added CheckDSDOldJourneyToggleAction to prevent users from accessing the old journey via URL when the feature switch is on


ajchilts (dismissed):
Approved . LGTM. Feel like I have already reviewed a lot of this in prior PRs. :-) 

👍 

Well done @matthew-goh

ajchilts (approved):

PR #363, opened 2025-05-13
ADR-1953 Remove stray </div> tag from MultipleSPRListView

matthew-goh:
Screenshot - 'page not working' link aligned with rest of content and further below the button, same as other pages
ADR-1953 after edit


ajchilts (approved):
That is a stray div alright!

PR #364, opened 2025-05-14
ADR-2025 Mandatory 2 decimal places for total litres in declaration and adjustment journeys

matthew-goh:
Content change for these forms: Fieldset labels include "(to [2/4] decimal places)" at the end; hint text removed.

Agreed with Wandi that where there were differences in the fieldset labels (e.g. Total litres of [beer], Litres of pure alcohol) between Confluence and the existing service, we should keep the labels as they are on this ticket, other than adding "(to [2/4] decimal places)".


app/views/adjustment/AdjustmentVolumeWithSPRView.scala.html
```diff
@@ -93,8 +91,9 @@
               )
               .asNumeric()
               .withWidth(Fixed10)
-              .withPoundPrefix
               .withHint(HintViewModel(messages("adjustmentSmallProducerReliefDutyRate.hint")))
+              .withPoundPrefix
+              .withSuffix(PrefixOrSuffix(content = messages("site.rate.litre")))
```
matthew-goh, line 96: "per litre" suffix added to match single SPR and multiple SPR forms

app/views/declareDuty/TellUsAboutSingleSPRRateView.scala.html
```diff
@@ -96,6 +94,7 @@
                 )
                 .asNumeric()
                 .withWidth(Fixed10)
+                .withHint(Hint(content = messages("tellUsAboutMultipleSPRRate.hint.currency")))
```
matthew-goh, line 97: Added missing hint text (shown on Confluence) to match multiple SPR and adjustment volume with SPR forms

ajchilts (dismissed):
LGTM @matthew-goh 👍

PR #366, opened 2025-05-15
ADR-2009 fix flaky tests in RichJsValueSpec

simondrugan16:
This PR is for the ticket [ADR-2009](https://jira.tools.tax.service.gov.uk/browse/ADR-2009) Fix flaky RichJsValueSpec test

The reason for "must return an error given a keyPathNode for an array" test to fail is because sometimes the sample returned none. Using retryUntil, rather than suchThat here fixed the issue since the generated value was never none.

The second flaky test "must remove a value given a keyPathNode and return the new object" occurred much more infrequently, however was caused, I believe, by the key chosen to be added already being part of the list, causing an error when processing. This was fixed by forcing the key to be added not already being a key in the JSObject.


ajchilts (approved):
Nice :thumbnail

PR #370, opened 2025-05-16
ADR-1965 Live Service Checklist - remove logic from views (attempt 2)

simondrugan16:
This is a PR for the ticket [ADR-1965](https://jira.tools.tax.service.gov.uk/browse/ADR-1965) Live Service Checklist - Investigate Returns FE use of logic in views.

I am searching for logic in the view that I can move to view models or controllers. This is good because logic in view files is an anti-pattern since the logic is not as well tested as other parts of the code.


test/controllers/adjustment/WhenDidYouPayDutyControllerSpec.scala
```diff
@@ -249,16 +282,23 @@ class WhenDidYouPayDutyControllerSpec extends SpecBase {
...
       }
     }
 
     "must redirect to Journey Recovery for a GET if no existing data is found" in {
+      val mockHelper = mock[WhenDidYouPayDutyHelper]
 
-      val application = applicationBuilder(userAnswers = None).build()
+      val application = applicationBuilder(userAnswers = None)
+        .overrides(bind[WhenDidYouPayDutyHelper].toInstance(mockHelper))
+        .build()
```
matthew-goh, line 301: For this test and those below, WhenDidYouPayDutyHelper is not called, so we don't have to create `mockHelper` and add it to `application`?

app/views/adjustment/WhenDidYouPayDutyView.scala.html
```diff
@@ -54,36 +43,30 @@
...
-        </p>
+            <p class="@Css.bodyCssClass">
+                @link(
+                    id = "excise-enquiries-link",
+                    text = messages("whenDidYouPayDuty.exciseEnquiries.link"),
+                    href = viewModel.exciseEnquiriesUrl,
+                    newTab = true
+                )
+                @messages("whenDidYouPayDuty.paragraph")
+            </p>
```
matthew-goh, line 60: Should we update this to use @paragraph instead of using <p> directly?

app/views/adjustment/SpoiltAlcoholicProductTypeView.scala.html
```diff
@@ -31,14 +31,14 @@
     formGroupWithError: FormGroupWithError
```
matthew-goh, line 31: Looks like helper is no longer used in this view

matthew-goh:
This is a small thing, but the names of the new tests use "should" instead of "must" 😆

ajchilts (approved):
LGTM 👍

test/controllers/adjustment/WhenDidYouPayDutyControllerSpec.scala
simondrugan16: i'm sorry i don't know what you mean

app/views/adjustment/WhenDidYouPayDutyView.scala.html
simondrugan16: this was autoformatter so not new code, so maybe a new ticket for this? maybe there was an issue putting a link in the paragraph template? dunno

app/views/adjustment/SpoiltAlcoholicProductTypeView.scala.html
simondrugan16: removed good spot

simondrugan16:
> This is a small thing, but the names of the new tests use "should" instead of "must" 😆

aah gosh darn must matchers 😂 I caught myself trying to use it on like every spec but clearly didn't notice myself using them :D. It was a dark day for me when they decided to use must matchers not should matchers 😢💀

app/views/adjustment/WhenDidYouPayDutyView.scala.html
matthew-goh: Ok, there could be another ticket to look for html in views. This one was probably because there's both a link and a normal message within the paragraph, but I think there are other places where it uses `HtmlContent` to include both

test/controllers/adjustment/WhenDidYouPayDutyControllerSpec.scala
matthew-goh: In these tests, the controller action redirects to journey recovery, so we don't reach the code that calls WhenDidYouPayDutyHelper. The tests still pass if we remove these 2 lines:
```
val mockHelper = mock[WhenDidYouPayDutyHelper]

.overrides(bind[WhenDidYouPayDutyHelper].toInstance(mockHelper))
```

app/views/adjustment/WhenDidYouPayDutyView.scala.html
ajchilts: I am not so concerned about this one. Probs OK as is.

test/controllers/adjustment/WhenDidYouPayDutyControllerSpec.scala
matthew-goh: Noted that it's better practice still to mock things that are injected but not called, even though other controller specs don't currently do this 🙂

matthew-goh (approved):

PR #371, opened 2025-05-21
ADR-1921: Fix content on a page in the spirits journey

Bradleigh-L:


matthew-goh (approved):
Although the updated message is now not consistent with the other regimes due to the addition of "(SPR)" in brackets and decapitalisation of "alcohol by volume", there is another ticket ADR-1860 in next sprint to change this on all pages.
Wandi has thus suggested that we go ahead with the change here.

PR #375, opened 2025-05-23
ADR-1955 Increase space between task list and BTA link

matthew-goh:
Before:
ADR-1955 before

After:
ADR-1955 after


matthew-goh:
PR bot comments addressed on ADR-1600

simondrugan16 (approved):

PR #376, opened 2025-05-28
ADR-1574 Return submitted page enhancements

matthew-goh:
From the 26th to the end of the month, we must time travel to the next month to test the scenario where payment is not overdue

### DIFF
--- a/app/views/checkAndSubmit/ReturnSubmittedView.scala.html
+++ b/app/views/checkAndSubmit/ReturnSubmittedView.scala.html
@@ -67,13 +67,9 @@
 
         @subHeading(messages("returnSubmitted.whatHappensNext"))
 
-        @if(viewModel.isPaymentOverdue) {
-            @paragraph(messages("returnSubmitted.positive.p1.overdue", viewModel.formattedPaymentDueDate))
-        } else {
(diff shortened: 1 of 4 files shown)


app/views/checkAndSubmit/ReturnSubmittedView.scala.html
```diff
@@ -57,17 +57,22 @@
...
                 link(
                     id = "view-return-link",
                     text = messages("returnSubmitted.viewReturnLink.text"),
                     href = controllers.returns.routes.ViewReturnController.onPageLoad(viewModel.periodKey).url
         )))
 
-        @subHeading(messages("returnSubmitted.positive.h1"))
+        @subHeading(messages("returnSubmitted.whatHappensNext"))
+
+        @if(viewModel.isPaymentOverdue) {
```
ajchilts, line 70: As this is logic should it live in the view model? I know we just did a load of work to move any logic into view models.

matthew-goh: In this case, because it's just one message, yes, the whole message could be passed in through the view model.

My thinking had been influenced by `@if(viewModel.returnDetails.isAmountGreaterThanZero)` in this view, as well as the logic in ViewPastPaymentsView, both of which have already been refactored and involve checking a Boolean value in the view model.

ajchilts (approved):
LGTM 👍

PR #389, opened 2025-06-13
ADR-2093: Focus status for Page not working properly link is not consistent

Bradleigh-L:


app/views/templates/Layout.scala.html
```diff
@@ -93,7 +93,7 @@
 @content = {
     @contentBlock
 
-    <div class="@Css.marginTop8CssClass">
+    <div class="govuk-!-margin-bottom-1 govuk-!-margin-top-8">
```
ajchilts, line 96: Can you please add this to the Css constants class?

ajchilts (changes requested):

ajchilts (approved):

PR #391, opened 2025-06-19
ADR-2072: Add link "Go back to view all tasks" on Return Journey

Bradleigh-L:


app/config/Constants.scala
```diff
@@ -26,6 +26,20 @@ object Constants {
 
   val noDetailsValue: String = "no-details"
 
+  val excludedBackLinkPaths: Set[String] = Set(
```
matthew-goh, line 29: I wonder if we should call these e.g. excludedReturnToTaskListPaths to avoid confusion with the actual back link

matthew-goh:
I also wonder if it's bette to have another Boolean parameter for the task list link here in the Layout instead of the list of paths in Constants:
`@(pageTitle: String, showBackLink: Boolean = true, ...)`
Then on each view where we don't want the link, we can put e.g. `showTaskListLink = false`, which should be clear and easy to maintain

app/views/templates/Layout.scala.html
```diff
@@ -15,6 +15,7 @@
  *@
 
 @import config.FrontendAppConfig
+@import config.Constants
```
matthew-goh, line 18: This import is no longer needed?

conf/messages.cy
```diff
@@ -1245,4 +1245,6 @@ returnSubmitted.noDetails.viewPayments = gwirio faint o doll sydd arnoch
 returnSubmitted.noDetails.p2 = Gall y manylion a nodwyd gennych yn y datganiad hwn gymryd hyd at 2 ddiwrnod gwaith i gael eu diweddaru yn ein gwasanaeth.
 returnSubmitted.noDetails.p3 = Defnyddiwch eich cyfeirnod o’r datganiad hwn fel cyfeirnod talu os oes angen i chi dalu unrhyw doll.
 
+viewAllTasks.Link.text = Ewch yn ôl i fwrw golwg dros yr holl dasgau
```
matthew-goh, line 1248: Maybe we should put this message somewhere near the task list messages (around line 245) instead of in the Check and submit sub-journey?

app/views/components/ReturnToTaskList.scala.html
```diff
@@ -0,0 +1,28 @@
...
+@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcLayout
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.language.{En, Cy}
+
+@this()
+
+@()(implicit messages: Messages)
+
+<p class="govuk-body">
+ <a href="@TaskListController.onPageLoad.url" class="govuk-link govuk-link--no-visited-state">@messages("viewAllTasks.Link.text")</a>
+</p>
```
matthew-goh, line 28: Perhaps this can use `linkNoVisitedStateCssClass`, or even the `Paragraph` and `Link` components?

matthew-goh:
So the link hasn't been added to MultipleSPRListView, AdjustmentListView and AdjustmentRepackagedTaxTypeView - is this intended?
Noted that AdjustmentRepackagedTaxTypeView will be removed when we make the changes to auto-populate the repackaged tax type

matthew-goh (approved):

matthew-goh:
Contained in ADR-2135, which has been merged

PR #394, opened 2025-06-23
ADR-2110 Remove new DSD journey and toggle

matthew-goh:
Remove all code used to implement the new DSD journey.

`cleanup` has been implemented for the initial yes/no question of the old journey - if the user selects no, subsequent pages in the journey are cleared as part of `UserAnswers .set()` instead of in a separate controller method. (`cleanup` was introduced with the new journey but hadn't been done for the old journey as we thought it would be removed.)


matthew-goh:
Version update has been done on ADR-1990

ajchilts (approved):
Nice 👍

PR #400, opened 2025-07-02
ADR-2138 Prevent access to CYA and Duty due pages if user answers have missing declarations

matthew-goh:
When the user lands on CYA or Duty due, CheckYourAnswersSummaryListHelper does a check to ensure that declaration details are present for the selected rate bands, as described in the ticket scope. If the check fails, the user is redirected to journey recovery.


ajchilts (approved):
Nice 👍

PR #401, opened 2025-07-03
ADR-2109 Content changes - 'What return period do you need to adjust?'

matthew-goh:
Various content changes to `/complete-return/adjustments/adjustment/declare/return-period` that make it clear that the user should input the return period being adjusted.

Files have been renamed from WhenDidYouPayDuty... to AdjustmentReturnPeriod...; similarly for messages.

The helper and view model for this page are no longer needed as (i) there is now a single title and heading for all adjustment types that show this page, and (ii) the guidance link url is in the messages file to pick up the right language.

The form for inputting the return period also has a new error message for invalid characters, separate from invalid numbers.

ADR-2109 English
ADR-2109 Welsh


app/pages/adjustment/AdjustmentReturnPeriodPage.scala
```diff
@@ -21,9 +21,9 @@ import play.api.libs.json.JsPath
 
 import java.time.YearMonth
 
-case object WhenDidYouPayDutyPage extends QuestionPage[YearMonth] {
+case object AdjustmentReturnPeriodPage extends QuestionPage[YearMonth] {
 
   override def path: JsPath = JsPath \ toString
 
-  override def toString: String = "whenDidYouPayDuty"
+  override def toString: String = "adjustmentReturnPeriod"
```
matthew-goh, line 28: This particular page isn't used to save data to user answers, so it's ok to change this

PR #404, opened 2025-07-09
ADR-2125: Update adjustment journey as per Design Standard Rate Tax Type applicable to Draught and Non draught

Bradleigh-L:


Bradleigh-L:
Screenshot 2025-07-09 at 16 09 50
Screenshot 2025-07-09 at 16 10 03

ajchilts (changes requested):
@Bradleigh-L  - I think we have a few issues here... 

Firstly the `toDescription` function is called from a lot of view models. Can we be sure we have only applied this to the pages specified on the ticket? 

Screenshot 2025-07-09 at 16 47 27


Secondly there are no tests for this PR. How do we assert the changes have been made and only on he pages specified on the ticket? 

I will spend a but of time looking into this one tomorrow.

matthew-goh (approved):

PR #414, opened 2025-08-01
ADR-2168: Accessibility Issue for new pages

Bradleigh-L:


Bradleigh-L:
Screenshot 2025-08-01 at 14 06 32
Screenshot 2025-08-01 at 14 05 58

app/views/declareDuty/MultipleSPRMissingDetailsConfirmationView.scala.html
```diff
@@ -32,6 +32,16 @@
 @(form: Form[_], regime: AlcoholRegime, missingRateBandDescriptions: Seq[HtmlContent])(implicit request: Request[_], messages: Messages)
 
 @layout(pageTitle = title(form, messages("multipleSPRMissingDetailsConfirmation.title"))) {
```
ajchilts, line 34: @Bradleigh-L  - Is this the only page that suffers this problem? I was under the impression there may be other pages? 

I am also not sure I like the style attribute directly in the view and wonder if we can move this to a css class? 

I can help look into this on Monday.

app/views/dutySuspended/DeclareDutySuspendedDeliveriesQuestionView.scala.html
```diff
@@ -33,8 +33,7 @@
 @(form: Form[_], mode: Mode)(implicit request: Request[_], messages: Messages)
 
 @layout(pageTitle = title(form, messages("declareDutySuspendedDeliveriesQuestion.title"))) {
-@Html(Css.legendFix)
-
+<div class="legend-fix">
```
ajchilts, line 36: There should not be a requirement to add this to a div. The CSS can be imported in the template.

ajchilts (changes requested):

PR #425, opened 2025-09-15
ADR-2251: Update Past payments page to allow customer to return to Pa…

Bradleigh-L:
…yment page

### DIFF
--- a/app/views/components/BackToPayments.scala.html
+++ b/app/views/components/BackToPayments.scala.html
@@ -1,0 +1,33 @@
+@*
+ * Copyright 2025 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


Bradleigh-L:
Screenshot 2025-09-15 at 10 16 47

app/views/components/BackToPayments.scala.html
```diff
@@ -0,0 +1,33 @@
...
+
+@()(implicit messages: Messages)
+
+@paragraph(content = HtmlContent(
+  link(
+        id = "back-to-payments-link",
+        text = messages("backToPayments.Link.text"),
+        href = controllers.payments.routes.ViewPastPaymentsController.onPageLoad.url
+    )
+))
```
matthew-goh, line 33: Creating this component is not a bad idea in case we later use it on other pages stemming from View payments!
But I noticed that the link isn't wrapped in a paragraph on the prototype, which is why the text is smaller in the screenshot 😄

PR #466, opened 2026-05-21
DDCE-8147 - ADR: Add a test script

tolgatunaaa:


lmugnai:
@tolgatunaaa we should also update build.sbt with 

addCommandAlias("scalafmtAll", "all scalafmtSbt scalafmt Test/scalafmt it/Test/scalafmt")

creates a new alias for scalafmtAll, overriding the task in sbt-scalafmt. Also keeps it consistent with all other repos

lmugnai (approved):

